Issue #51, opened 2016-08-31
MPI data preparation


pierrethibault:
Data preparation sometime fails when running with MPI. We should either debug this or consider falling back to single-process data preparation.

alexbjorling:
I'm having the issue that _mpi_save_chunk() fails to MPI-send data and instead raises the ValueError "ndarray is not contiguous". It works if you modify utils.parallel.send() such that it sends the contiguous copy np.ascontiguousarray(data) of the array, but I guess that this is not the proper solution.

Is this the failure you are referring to, @pierrethibault ?

pierrethibault:
@stefanoschalkidis Can you confirm that this is also the error message you get? If so, I think that making a contiguous copy is a good fix for now.

stefanoschalkidis:
I'm currently getting a different .h5 error, but this is probably due to the messy structure of the I13 subclass. I will try to confirm the above once I've reorganized I13.

alexbjorling:
I'm getting this error both with some I13 far-field data that Aaron kindly sent me, and with data taken here at NanoMax and loaded with my own PtyScan subclass. It happens as soon as I run on more than one process.

stefanoschalkidis:
Yes, I think I got a similar error with I13 near-field data, but then I introduced some switches to the subclass and now the code breaks earlier due to an .h5 error when run with more than one core. I intend to reorganize the script for near-field soon and will then try to recreate the original behaviour.

stefanoschalkidis:
Hi all, I'm done with reorganising the I13 NFP subclass and now I don't get any error when loading with MPI. The code however breaks later in classes.py, but the prepared .ptyd file is perfectly fine, meaning it will work when the reconstruction is loaded directly from there (also with MPI). I will try to look into the new error, but for now I don't think I can comment on the _mpi_save_chunk() failure.

stefanoschalkidis:
Hi, out of curiosity I also reorganised the I13 FFP subclass and now I get the same error. The contiguous array fixes this issue, but then the code stops at the hard coded megapixel limit. @alexbjorling do you observe the same behaviour?

alexbjorling:
No I don't, but maybe you have a larger dataset than I do.

@pierrethibault, is this the same error that you refer to in the original issue?

pierrethibault:
[...]

PR #71, opened 2016-11-30
Repairing the data array after cropping/flipping etc.

alexbjorling:
This fixes MPI data preparation bug #51, by restoring contiguity of data arrays after cropping/padding/binning/flipping/rotating. Please test!

### DIFF
--- a/ptypy/core/data.py
+++ b/ptypy/core/data.py
@@ -725,6 +725,9 @@ class PtyScan(object):
                         'Binning (%d) is to large or incompatible with array '
                         'shape (%s).' % (rebin, str(tuple(sh))))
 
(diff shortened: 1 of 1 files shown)


bjoernenders (approved):

simonesala:
From a quick inspection, after this commit MPI still does not work for me.
More specifically I get KeyErrors from core/data.py while running on my FERMI_SS branch (using experiment/DiProI_FERMI.py) and I get the known RuntimeError 'Arrays larger than 50M not supported' while running on the master branch on data that still works nicely when I do non-parallel preparation followed by parallel reconstruction. Any suggestions?
I'll look more into this.

Issue #49, opened 2016-08-30
Multi-pass model construction 


pierrethibault:
The model construction currently is a nested loop very difficult to update. A more flexible option would be to have a multi-pass approach that starts with a "vanilla pod" and gradually builds the model depending on input parameters. A new feature in the model would be added as an additional pass.

bjoernenders:
I thought about this tonight and I think a relational database might be useful. Each pod will have an entry, at least those inferred from the Ptyscan, "Scan no", "Diff index","Postion" etc. in the model creation phase we add more characteristics or update those, like probe_mode_index etc. The view creation can then be deducted from the record, following a set of rules that is taylored to the application  bragg / multislice / opr etc.  At this point the record info will be used to determine the accessrules.

The big bonus of a relational database is, that it is very easy to group pods with a common attribute. Such way we can create new records with the help of records allready in the database. For example get all records with probe_mode_index = 0 and add create additional records of the same kind with probe_mode_index = 1. If new records are created, this action will trigger view and pod creation etc.  Such things may be super hard to do in python. And we could use a light-weight database like sqlite. 
The fun part is that this kind of data base is transparent when it come to new data. Because when a new datapoint comes in one has to just look for all records of the last data point and create as many records with the new datapoint. 

The only concern is the speed of the database, but our current pod creation mechanism is superslow anyway.

pierrethibault:
I very very much like the idea. As far as I can tell more and more "big data" science is done with relational databases so it seems quite natural to drift toward this option. I would be surprise if speed was a big problem.

aaron-parsons:
Ok time for me to ask some stupid questions here. I like the idea of the database, but want to be clear I'm thinking of the same thing as you guys.
1. Do we want this to be a database on disc or in RAM?
2. Would pytables ([https://www.pytables.org/usersguide/tutorials.html]))do the job? (To me it seems like it might be quite easy to bolt on to the scan model).

aaron-parsons:
This already sketched using the recarray, but would need more thought going forward.

PR #73, opened 2017-01-11
Generalizing container and geometry classes to 3d

alexbjorling:
I'm going for 3D Bragg ptychography and would like to do this within the ptypy framework. As a first step, the container classes need generalizing.

I'm not sure if this makes any sense from a ptypy perspective, but the generalization seems pretty unintrusive and doesn't change (for example) the memory layout for 2D containers.

I've tested as far as I've been able to, for example by going through the ptypyclasses.py tutorial and making sure everything is as expected. I'm sure more testing would be needed to make sure we don't create bugs in the heart of the library...

Consider this a starting point for discussion.


alexbjorling:
Ok, @bjoernenders and @pierrethibault, I've made some headway on Bragg 3d ptycho here. 

What I have is container classes and a geometry class which describe the 3d rocking curve experiment. I've then used simple scripts (see `tutorials/bragg3d_initial.py`) to manually create and keep track of the relevant `View` objects, so no `POD`s yet. We've discussed how to reorganize model construction, and obviously I'd like to create models and `POD`s in the long run to handle incoming data, parallelization etc. But I've started at the lower level.

What I've done is to _change_ the `Container`, `Storage`, and `View` classes slightly, adding a dimensionality parameter which can be 2 or 3. On the other hand, since the geometry is so different I have instead _inherited_ the `Geo` class to keep some of its nice updating functionality, to a subclass called `Geo_Bragg`. Perhaps these two should instead have a common parent? 

What do you think of this approach in general?

bjoernenders:
Hej Alex,
congrats on your progress! I like the coding style. I thought I already commented here, but I don't find my original comments. I browsed through the files and here are a few of my thoughts:

1) The additional dimension, the 'aisles' are added as last dimension. In memory layout, this would correspond to the fast axis. Personally I think this axis belongs in between the 'layer' and the former 2d view axes. In the way you implemented it now, its a dual system with 2d and 3d side-by-side. We could think of a more general system that starts of with 4d Storages any way, and both 2d and 3d view access. That could be somehow beneficial. Originally I was hoping you would utilize the arbitrary layer access to get 3d views. My general point is that the position of the added dimension appears to me somewhat non-ideal in view of a converged platform. What is your opinion?

2) In order to allow for 2 systems to coexist, you implemented checks and if-switches. Have you noticed a performance drawback when using 2d views when compared to the old classes.py? The problem is that certain types of if-switches can delay the program significantly, as they open a new contexts. So especially for the final data access we should try to avoid it. I didn't benchmark this.

3) Bragg ptycho is not likely to be used in the same scenario as conventional ptycho. Personally I would be in favor of a global context switch. Thus, when ptypy is imported, it knows already if what you want. Like in __init__ it loads a config file and determines from that point what other modules to load. We already thought about such a possibilty and my opinion is, it is a must-have. Then you could place all the files you need in a subfolder "/bragg" for now and sideload its modules when a relevant config parameter indicates that. 

Just a few thoughts.

alexbjorling:
Thanks Björn for your comments!

1) You're right, the new axis should definitely be the slowest running in memory, so that it goes "pages, rows, columns". I think right now this interpretation of the axes is just a matter of semantics in Storage.reformat(), but I might be wrong. In any case, my feeling is that it would be best to always have 4d Storages, and to let 2d Views squeeze away the "pages" dimension. As long as the 2d case uses the last two dimensions, it should be OK memory wise. 

2) Good point. I actually went for the if switches specifically to avoid performance loss in the old Storages from unintentionally changing the memory layout. I'm leaning towards a more general 4d solution as above, but I'll do some performance tests comparing the current version with master, just to know.

3) Actually, in Bragg ptycho you will need a bunch of 2d storages as well. The experiment starts with finding the probe through 2d transmission ptycho of a test object. This can be done in a separate ptypy process of course, but yields a 2d storage that you pass to the Geo_Bragg object for extrusion and conversion into the right coordinate system. Another scenario is simultaneously using a 2d detector in the forward direction, for which you want another 2d Storage per position (and another 2d diff view attached to the POD). So, you need 2d and 3d storages side-by-side, but perhaps this is still possible with the different contexts you describe.

Let me know how you're thinking about the 4d generalization, if you like that idea then I can implement it here in the PR and find out if it's OK performance wise.

alexbjorling:
Ok, did some speed checks where I create, write to, and read from a large number of `View`s on a large complex `Storage`. Leaving the code out to avoid cluttering too much here.

With the master branch code,
~~~
...
creating a 1054x1054 complex storage with 10000 views of 64x64: 3.472 s
filling those views: 0.087 s
reading those views: 0.032 s
creating a 1054x1054 complex storage with 10000 views of 64x64: 3.466 s
filling those views: 0.088 s
reading those views: 0.031 s
creating a 1054x1054 complex storage with 10000 views of 64x64: 3.461 s
filling those views: 0.087 s
reading those views: 0.032 s
*****
Averages over 20 runs:
Total storage+view creation: 3.516366 s +- 0.006122 s
Total writing: 0.089317 s +- 0.000011 s
Total reading: 0.031841 s +- 0.000001 s
~~~

and with my side-by-side 2d/3d code, using `data_dims=2`:

~~~
...
creating a 1054x1054 complex storage with 10000 views of 64x64: 3.370 s
filling those views: 0.086 s
reading those views: 0.032 s
creating a 1054x1054 complex storage with 10000 views of 64x64: 3.371 s
filling those views: 0.086 s
reading those views: 0.032 s
creating a 1054x1054 complex storage with 10000 views of 64x64: 3.376 s
filling those views: 0.086 s
reading those views: 0.032 s
*****
Averages over 20 runs:
Total storage+view creation: 3.403871 s +- 0.003094 s
Total writing: 0.086351 s +- 0.000008 s
Total reading: 0.032065 s +- 0.000001 s
~~~

So my conclusion would be that the 2d/3d implementation takes a tiny bit longer when setting up storages and views, but that writing and reading to the buffers is not affected. So now I'm thinking this might be an acceptable solution and that generalizing to 3d (+ layers) always could be overkill.

bjoernenders:
Thank you Alex, for checking the performance. And sorry for my late reply, I was on vacation.
Ok I guess we keep your solution for now. I will follow the progress on your branch. Once the Bragg thing works, we discuss the merge. Personally I think we should consider some face time for that too. 
Good luck. :)

alexbjorling:
I've now managed to reconstruct both numerical and experimental data. I've done this in standalone scripts which simply use the generalized `Container`s and the new `Geo_Bragg` class, with manual loading of data, manual `View` construction, and explicit loops over `View`s for refinement. Keeping track of geometrical Bragg weirdness and using the 3d `Container`s for managing `View`s through ptypy is a huge help and dramatically simplifies data handling and refinement, even without using the `PtyScan` and `POD` logic. I've also used the `Geo_Bragg` class for Bragg CDI which has been very helpful.

I'm now unsure of how to proceed, and see a number of options:

(i) To request a merge of this pull request, and to keep actual Bragg refinement manual as it is now. Ptypy would then be a huge support but not a full solution to Bragg ptycho analysis. This is the simplest way but would not benefit from things like the ptypy parallelization and `POD` scheme.

(ii) To fork ptypy and adapt `PtyScan` and `POD` creation in an ad hoc way, which would avoid weighing down ptypy-proper, avoid complicated model construction, but also make the Bragg fork a sort of dead end which doesn't benefit from future development.

(iii) To engage (hopefully together with @pierrethibault, @bjoernenders and others) in proper model construction, ultimately making Bragg ptycho a part of actual ptypy.

What do you think?

aaron-parsons:
For what it's worth, my opinion would be to include it in the main package until such time as it starts to cause too many issues. This way the "forks" don't diverge into something incompatible. It's also very useful as an example use case for other people who are developing, on how to further the structures that already exist. 

The refinement step could be factored as a separate package inside the ptypy repo, with a view to a further refactor down the line as people start to use it/ add things.

Sadly, I can't offer much in the way of effort in developing this feature for a while, but I would certainly be keen to use it! Nice work.

pierrethibault:
I agree that it would be better to merge the changes while they are still compatible with the main package. Alex's option iii) is certainly the best since POD creation management is something that needs to be rewrote in the main code as well (see #49 ). So, better designing the update while integrating Bragg ptycho from the beginning. Björn's idea of using a light-weight database is very appealing to me, though I must admit I don't know much how this works.

alexbjorling:
I'm really glad to hear this, I would also like to have it integrated. The thing is I really don't know where to start, especially since the integration will have to coincide with the general rewrite of the main code.

alexbjorling:
Closing this (see PR #113 )

PR #83, opened 2017-06-03
Unittests for classes.py

stefanoschalkidis:


bjoernenders:
Hi Stefano, 
wow I see you come a long way with Python. It's a bit to detailed for us to be useful, since the classes are not yet a fixed structure. Let's talk about it at the workshop.
Bjoern

pierrethibault:
Stefanos won't be at the workshop. 

I don't really agree Björn: unit tests cannot be too detailed: they map as closely as possible to the current implementation, and can actually reveal how refactoring should be done. Adept of test-driven development, as Stefanos has become, would tell you that this is the right way to do it: write the tests *before* you do any change. Since this is a single new file and that the travis tests have passed I would be inclined to merge right away.

bjoernenders:
Ok I agree it does no harm. And I totally agree that Stefanos has become adept and that this is a well written test in terms of coverage and great work. But in my defense, the detailed test case here tests the pathological single pixel and a pixel size of 1 and doesn't test the access pattern which is the important part that we do not want to break. But I just realized, I did overlook Stefanos comment in the issue tracker claiming that this is the start. It is a good start indeed.

bjoernenders (changes requested):
Please define immutables like 'S0000' and (1,1,1) that are used repeatedly and "setup" the test arena at the beginning of the script.

stefanoschalkidis:
Thank you for your feedback. I know it may seem too precise at times, but it is actually quite challenging to write meaningful unittests to existing code. The current implementation will be refined over time as the code evolves. For new features the unittests should be written in advance and subsequently the code modified until it passes the tests.

I’m a bit busy with data analysis at the moment but hope to continue work on the unittests as soon as I find some free time.

stefanoschalkidis:
Hi Björn,

I’ve added the test arena to the beginning of the script. Let me know if it needs to be more elaborate. I haven’t changed the immutables as I believe the tests should be as self-contained and independent as possible.

Kind regards,
Stefanos

bjoernenders (approved):

bjoernenders:
Hi Stefanos,
Thanks for all the work. There is still the issue that this setup tests the pathological case of a single pixel held in Storages or Containers with pixel size 1. I was more thinking along the lines that we could have different Storage/ View shapes and pixel sizes. But maybe that is a stretch goal for the future.
Thanks again,
Bjoern

stefanoschalkidis:
Hi Björn,

You're welcome! I completely agree that some of the tests are quite basic. My idea was to catch the cases you're mentioning with a test from user's perspective simply following the examples in the documentation. Let me know what you think about this.

Btw, are the tests working for everyone?

PR #86, opened 2017-07-11
Some changes to Ptypy

blochl:
This series introduces some changes to Ptypy, the most important of which is using pyFFTW as the default method to perform the Fourier transforms, a change which boosts the performance by up to 60%.
Besides that, some minor optimizations are made to slightly improve the performance beyond that, as well as some refactoring which carries no performance benefits, but makes the code more readable and PEP 8 compliant.

These patches are based on the currently available master branch (commit 208ca77) and are tested on a series of real-world NFP datasets.

Best regards,
Leonid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,6 @@ script:
 
 after_script:
   - coveralls
-  
+
 allow_failures:
 - python: "3.3"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ before_install:
   - conda update -q conda # update this quietly
   - conda install pyyaml
   - conda info -a # and print the info
-  - conda install pyyaml
 
 install:
   - conda env create --file environment.yml #set up the environment thats in the file
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,7 @@ dependencies:
   - pyzmq
   - pep8
   - pytest
-  - mpi4py 
+  - mpi4py
   - pil
   - pip:
     - pytest-cov
--- a/environment.yml
+++ b/environment.yml
@@ -1,4 +1,6 @@
 name: test-environment
+channels:
+  - conda-forge
 dependencies:
   - python=2.7
   - numpy
@@ -11,6 +13,7 @@ dependencies:
   - pytest
   - mpi4py
   - pil
+  - pyfftw
   - pip:
     - pytest-cov
     - coveralls
--- a/environment.yml
+++ b/environment.yml
@@ -17,5 +17,3 @@ dependencies:
   - pip:
     - pytest-cov
     - coveralls
-
-
--- a/ptypy/__init__.py
+++ b/ptypy/__init__.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
 # -*- coding: utf-8 -*-
 from version import short_version, version, release
 """
--- a/ptypy/__init__.py
+++ b/ptypy/__init__.py
@@ -6,7 +6,7 @@ PTYPY(v%(short)s): A ptychography reconstruction package.
 
 To cite PTYPY in publications, use
  @article{ ... }
- 
+
     :copyright: Copyright 2014 by the PTYPY team, see AUTHORS.
     :license: GPLv2, see LICENSE for details.
     :version: %(version)s
@@ -20,7 +20,7 @@ try:
     __has_zmq__= True
     del zmq
 except ImportError('ZeroMQ not found.\nInteraction server & client disabled.\n\
-Install python-zmq via the package repositories or with `pip install --user pyzmq`'): 
+Install python-zmq via the package repositories or with `pip install --user pyzmq`'):
     __has_zmq__= False
 
 try:
@@ -29,7 +29,7 @@ try:
     del mpi4py
 except ImportError('Message Passaging for Python (mpi4py) not found.\n\
 CPU-parallelization disabled.\n\
-Install python-mpi4py via the package repositories or with `pip install --user mpi4py`'): 
+Install python-mpi4py via the package repositories or with `pip install --user mpi4py`'):
     __has_mpi4py__= False
 
 try:
@@ -40,7 +40,7 @@ except ImportError('Plotting for Python (matplotlib) not found.\n\
 Plotting disabled.\n\
 Install python-matplotlib via the package repositories or with `pip install --user matplotlib`'):
     __has_matplotlib__= False
-       
+
 # Initialize MPI (eventually GPU)
 from utils import parallel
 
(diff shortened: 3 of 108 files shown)


pierrethibault:
Thanks for this. I am happy with most of the changes. I have one request and one question for @bjoernenders and @aaron-parsons.

The request: please modify `geometry.py` in such a way that `pyFFTW` is only an optional dependency, with fallback to `scipy`.

The question: can someone comment on the change in relative imports ([lines 20-36 of data.py](https://github.com/blochl/ptypy/blob/ec8acf5982303b94a3e473029edb782d54fe7b03/ptypy/core/data.py#L20-L36))? Is it ok like this?

bjoernenders:
Hello Leonid,
thanks for your commit. Happy to see a new contributor! I'll comment after the weekend, I am a bit behind my workload right now. Same for you @pierrethibault. Sorry
Best
Bjoern

blochl:
@pierrethibault : Thanks for your review! Now pyFFTW is an optional dependency, and if for some reason it is not available, a fallback to SciPy occurs. I also reduced code redundancy with this change.
@bjoernenders : Thank you! Yes, I'm quite busy these days as well.

Additional comments:
1) I discovered accidently, when testing on a limited-memory machine, that FFTW also reduces the required RAM for processing over SciPy FFT.
2) The new changes are tested on AWS EC2 instances (and do work as expected).

bjoernenders (commented):
Regarding f28fa23: Thank you for the idea of the automated whitespace removal. The sheer mass of changes is too much for me to review now. There may be whitespaces needed for the autodoc. So i don't approve of this patch until I see the documentation being rebuild properly. Since the doc moves into main very soon, I think this patch has to white until then. But the command may be very helpful then, thanks.

Regarding e1f5744, 89b1bbf, 77bcf6b: Fine with me
  
Regarding f125fe4: Even though the choice of the absolute square seems odd, we used one that works with only complex numbers for a reason. But I don't quite recall it now. The function was called "cabs2" because it is not returning the result as a real but as a complex in order not to mix it up with the general abs2. Yeah there could have been an error with the docstrings, but I personally don't think this function is critical for performance. So I don't like this commit, please remove it from the patch.

Regarding f82b9ba: When I tested it like 5 years ago, the Pythonic square was not faster for numpy arrays.

Regarding 04f4d8d: I don't mind using scipy there. "Which is known to be faster" ... I didn't know that specifically. What's your source?

Regarding e4e5fe4: The "speed" argument. I heard this argument a lot know. From each and every Master student. In the bread-and-butter case of ff ptycho it always fails to show in the end, so the external dependency and the plan making biz of pyfftw was never worth it to me. For the large non-cacheable frames of nf ptycho, pyfftw may finally found its application. Also installing pyfftw has its intricacies and fails sometimes. Since it is now optional, I'm ok with this one.

blochl:
@bjoernenders : Thanks for the review.
Regarding the trailing whitespace removal, instead of reviewing each change manually, you could run the commands yourself and compare the diff, to make sure that no other changes were introduced. As for interference with the docs, I can hardly think that trailing whitespaces would make a difference.

Absolute value: <code>cabs2</code> is used **only** in 2 places: for DM (**DM.py**, **DM_simple.py**) and in **engine_utils.py**. In **engine_utils.py** it is used as a real number only (<code>u.cabs2(...).real</code>) so it is **exactly** the same behaviour now. In the DM procedures, it is used indeed for an addition to a complex array, but it is multiplied by an integer before that (which makes things slower) and the type is casted to complex anyway, because you add it to a complex value later (and addition of real to complex has less overhead than multiplication of complex by real). That's it. There are no more uses of <code>cabs2</code> in the code. There may have been a reason to introduce it, but there is no reason to keep it. The performance enhancement is indeed very slight (but noticeable - ~1-2%) but this patch also removes code redundancy and unclarity, making the process more logical. The performance is a bonus.

bjoernenders:
You're right @blochl , it is used in this way `u.cabs2(..),real` for engine_utils.py and the other way is probably better. I overlooked that part, my bad.   
I would still like to keep the cabs2 in **DM.py** and **DM_simple.py** and the function `cabs2` itself. It was created for a reason. What you put it as an improvement, was the function we originally used. It changed to this someday as a result of bug hunting, which is why I want to keep it this way. Also we cannot just delete an entire function, since it may be used in dependent code. Even though this function is tiny.

blochl:
@bjoernenders : Thanks for your remarks.
* Yes, you are right, it is better not to remove <code>cabs2</code> completely, as someone may depend on it. I have also returned <code>cabs2</code> to DM procedures. It was introduced in a415a687d, with the statement "Manny many bugfuxes in incomplete defaults" [*sic*]. Since I do not know what these bugfixes are, I guess that indeed it is better to leave it there, as you have said.

  Performance comparison of the previous and the current <code>abs2</code>:
```python
In [1]: import numpy as np
In [2]: A = np.random.randn(2 * 1000**2).view('c16')
In [3]: A = A.reshape((1000, 1000))
In [4]: %timeit np.square(np.abs(A))
31.7 ms ± 34 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
In [5]: %timeit (A * A.conj()).real
7.56 ms ± 97.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

* As for Pythonic square vs. <code>np.square</code>: I do find a slight performance boost on one machine, but slight performance decrease on another machine. So I do not insist on this patch, and can remove it. To illustrate the order of magnitude:
```python
In [1]: import numpy as np
In [2]: A = np.random.randn(2 * 1000**2).view('c16').reshape((1000, 1000))
In [3]: %timeit A**2
3.56 ms ± 77.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [4]: %timeit np.square(A)
5.64 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

* Numpy <code>eig</code> vs. SciPy <code>eig</code>: Numpy uses LAPACK_lite, while SciPy uses the more complete LAPACK procedures. Also, the benchmarking is quite consistent:
```python
In [1]: import numpy as np
In [2]: from scipy.linalg import eig
In [3]: A = np.random.randn(100**2).reshape((100, 100))
In [4]: %timeit np.linalg.eig(A)
76 ms ± 13.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
In [5]: %timeit eig(A)
56.3 ms ± 10.8 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

* Importance of pyFFTW: I work on NFP datasets which previously took 28+ hours per projection to process on 16 cores, and there are 1000 projections in a single dataset. After introducing pyFFTW, it takes now 14-16 hours. So this is a **huge** boost. Also, on machines with 30 GB of RAM (c4.4xlarge instances on AWS) the processing is impossible with SciPy fftpack, but is made possible only by pyFFTW, because of RAM limitations.

* Why do I want to "squeeze" every drop of performance: It currently takes about a month to process just a single dataset, on tenths of 16-core machines. I'm also experimenting with processing on AWS, where each hour costs money. So even the slightest increase in performance will have a significant impact. I also experiment with Numba and GPU currently, but that's just experimental for now.

Issue #74, opened 2017-01-16
Possible bug in the View class


alexbjorling:
While working with my 3D `Storage`s and `View`s, I found some very weird behavior which I can also reproduce for 2D objects on the `master` branch.

For certain combinations of `psize` and `coord`, getting a view on a storage like `S[V]` returns nonsense. Reformatting the `Storage` alternately gives two incorrect views (see below). Are there some constraints on `psize`, `coord`, `shape`, etc, that I have missed? 

Code:
~~~
from ptypy.core import Container, View
import numpy as np

# The following works with psize = .2 or .05 but not .1, and with center = 54.7 or 54.8 but not 54.75
psize = .1
center = np.array([ 54.75,  54.75])
shape = (84,75)

C = ptypy.core.Container(data_dims=2)
S = C.new_storage(shape=128, padonly=False, psize=psize)
V = View(container=C, storageID=S.ID, coord=center, 
                     shape=shape, psize=psize)

for i in range(10):
    print 'S[V].shape = %s,'%str(S[V].shape)
    S.reformat()
print 'S.shape = %s'%str(S.shape)
~~~

Output:
~~~
S[V].shape = (0, 0),
S[V].shape = (0, 0),
S[V].shape = (84, 74),
S[V].shape = (84, 0),
S[V].shape = (84, 74),
S[V].shape = (84, 0),
S[V].shape = (84, 74),
S[V].shape = (84, 0),
S[V].shape = (84, 74),
S[V].shape = (84, 0),
S.shape = (1, 84, 75)

~~~

bjoernenders:
I confirm this bug. Thanks for reporting (Although the script above does not execute as ptypy is not imported).
A long time ago, it took me a while to stabilize this reformatting issue. I will investigate this and get back to you. Maybe this should be a part of unit testing too.

alexbjorling:
Bumping this bug (of which I have no memory!). Will also try to look into it.

alexbjorling:
OK the problem here is that if pcoord is right between two pixels, rounding becomes unpredictable and all goes bad. Filed a PR with a workaround.

PR #93, opened 2017-08-23
Workaround for the rounding errors that cause issue #74

alexbjorling:
Making sure that dcoord=round(pcoord) is always rounded in the same direction if the decimal part of pcoord is very close to 0.5. It's essential, because otherwise the behavior is unpredictable and Storage.reformat() screws up the Views as in issue #74.

I'm sure this can be done in a prettier way, using the decimal module or something. But this is simple and works.

### DIFF
--- a/ptypy/core/classes.py
+++ b/ptypy/core/classes.py
@@ -549,7 +549,7 @@ class Storage(Base):
 
         # Integer part (note that np.round is not stable for odd arrays)
         v.dcoord = np.empty((2,), dtype=int)
-        for i in (0, 1):
+        for i in range(len(v.pcoord)):
             dec = (v.pcoord[i] - np.floor(v.pcoord[i]))
             if dec > .4999 and dec < .5001:
                 v.dcoord[i] = int(np.ceil(v.pcoord[i]))
--- a/ptypy/core/classes.py
+++ b/ptypy/core/classes.py
@@ -548,13 +548,7 @@ class Storage(Base):
             v.pcoord = self._to_pix(v.coord)
 
         # Integer part (note that np.round is not stable for odd arrays)
-        v.dcoord = np.empty((2,), dtype=int)
-        for i in range(len(v.pcoord)):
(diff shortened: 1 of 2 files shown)


pierrethibault:
Are you sure this solves the issue? Doesn't this just move the problem from .50000000 to .49990000 ?

alexbjorling:
Good question. I think it solves it because below .4999 and above .5001 (or
whatever) np.round is predictable. The problem before was that it was
apparently going randomly in both directions when pcoord was assigned to
x.5 and therefore stored as x.500000000001 (or so).

At least for the example in the issue it behaves now.


Den 24 aug. 2017 15:14 skrev "Pierre Thibault" <notifications@github.com>:

Are you sure this solves the issue? Doesn't this just move the problem from
.50000000 to .49990000 ?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ptycho/ptypy/pull/93#issuecomment-324631121>, or mute
the thread

alexbjorling:
Oh ok now I see what you mean. The comparison (x > .4999) is then unpredictable as well. No then this doesn't fix it.

pierrethibault:
Ok I see. The problem is not `np.round` per se but the unfortunate fact that a value of exactly 0.5 is more likely than others because of manually entered parameters. So I agree with offsetting this by a small amount, though I think that 
`np.round(v.pcoord + 0.00001).astype(int)` would do the job just as well.

_Sent from my Samsung SM-A300FU using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

bjoernenders:
I agree to Pierres proposition. We could also think about disallowing a view to have odd number of pixels along the fast axes.

alexbjorling:
Funny that I spent so long looking for the problem that, when I found it, I completely misunderstood it. But yes, there is always going to be a problem somewhere, but having it at 0.5 is just really unfortunate.

bjoernenders (approved):

Issue #91, opened 2017-08-17
Working on windows?


alexbjorling:
One of my users is considering switching to ptypy. The problem is that he wants to do it on windows.

He's reporting 3 symptoms (see output)
* Everything that ptypy prints is printed 4 times (is it doing MPI by itself?). But stuff from the nanomax PtyScan is printed only once.
* Automatic COM guessing fails miserably and gives negative values
* After that, it crashes in a way suggesting that the imports don't work, saying that `logger` and `parallel` are `None`.

This behavior makes me feel that there's something windows-incompatible at least with the way imports are done. Does anyone know if the current code can run on windows at all?

~~~
loading data: scan_0192_pil100k_0016.hdf5
> loading normalization data: scan_0192_pil100k_0016.hdf5
> normalizing line by:  [
> 1.05667706  1.06190191  1.06141107  1.0556026   1.03927968  1.0199902
> 6
>  
> 0.99759464  0.9760873   0.95855387  0.95383338  0.96124094  0.9831006
> 6
>  
> 1.01826523  1.05484035  1.08091398  1.0938926   1.09300157  1.0904335
> 1
>   1.08815937  1.08427569  1.07589705]
> loaded 17 lines of Pilatus data
> ROI center is [ 98 238], automatic guess is [-27.08123249
> -12.78431373].
> ROI center is [ 98 238], automatic guess is [-27.08123249
> -12.78431373].
> ROI center is [ 98 238], automatic guess is [-27.08123249
> -12.78431373].
> ROI center is [ 98 238], automatic guess is [-27.08123249
> -12.78431373].
> Enter preprocessing (crop/pad True, rebin False, flip/rotate False)
> ... 
> 
> Enter preprocessing (crop/pad True, rebin False, flip/rotate False)
> ... 
> 
> Enter preprocessing (crop/pad True, rebin False, flip/rotate False)
> ... 
> 
> Enter preprocessing (crop/pad True, rebin False, flip/rotate False)
> ... 
> 
> Feeding data chunk
> Feeding data chunk
> Feeding data chunk
> Feeding data chunk
> Importing data from NM as scan NM.
> Importing data from NM as scan NM.
> Importing data from NM as scan NM.
> Importing data from NM as scan NM.
> End of scan reached
> End of scan reached
> End of scan reached
> End of scan reached
> End of scan reached
> End of scan reached
> End of scan reached
> End of scan reached
> C:\Users\dsf-user\AppData\Roaming\Python\Python27\site-
[...]

PR #95, opened 2017-08-30
Fixed windows COM overflow issue

alexbjorling:
Fixes the center automatic guess overflow issue in #91 for windows where `np.int32` seems to be the default `int`.

### DIFF
--- a/ptypy/utils/scripts.py
+++ b/ptypy/utils/scripts.py
@@ -522,7 +522,7 @@ def mass_center(A, axes=None):
     else:
         axes = tuple(np.array(axes) + 1)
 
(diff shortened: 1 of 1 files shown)


bjoernenders:
I don't think this is the right fix, because input array A can be a floating point value in general. I think it is best to cast `dtype = np.float` in both sums.

bjoernenders:
Could you verify if the bug is still fixed, Alex?

alexbjorling:
Yep still fixed.

You could also do `np.indices(A.shape, dtype=np.int64)` but I guess casting to floats gives the least risk of overflow somehow.

PR #96, opened 2017-08-31
Bugfix for data prepared prior to fly scan commit

stefanoschalkidis:
num_frames_actual was introduced with the fly scan commit, leading to a None TypeError when older prepared data is run with the current version of the code

### DIFF
--- a/ptypy/core/data.py
+++ b/ptypy/core/data.py
@@ -1204,6 +1204,8 @@ class PtydScan(PtyScan):
         with h5py.File(source, 'r') as f:
             check = f.get('chunks/0')
             # Get number of frames supposedly in the file
+            # FIXME: try/except clause only for backward compatibilty 
+            # for .ptyd files created priot to commit 2e626ff
             try:
                 source_frames = f.get('info/num_frames_actual')[...].item()
             except TypeError:
--- a/ptypy/core/data.py
+++ b/ptypy/core/data.py
@@ -1204,7 +1204,10 @@ class PtydScan(PtyScan):
         with h5py.File(source, 'r') as f:
             check = f.get('chunks/0')
             # Get number of frames supposedly in the file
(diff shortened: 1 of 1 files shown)


bjoernenders:
Thanks for the fix, Stefanos.
I am not in favor of fixes for backward compatibility. If we place them, there should be an annotation or a warning, such that we easily find them later. I put an annotation in for now. 

I think a better strategy is to update your files with a script and share that script in the /scripts section.   This is a general issue how to deal with backward compatibility, maybe @pierrethibault, @alexbjorling or @aaron-parsons  want to comment on it?

alexbjorling:
Other programs (like gromacs) increment version counters on their internal binary file formats when they change, to let the reader know what to expect.

I guess if the ptyd/ptyr files don't have version numbers then backward compatibility is too ambitious anyway. If the format evolves (like with ScanModels or what not) then that will generate a lot of try/catch messiness. 

On the other hand, if you use old ptyr files as input, it would be really annoying to be forced to recreate those just because you upgraded ptypy.

aaron-parsons:
Yes this is true with NeXus and CXI too. We did talk about moving away from the ptyr and ptyd formats towards a standard, did that discussion go any further?

Alternatively, we could accept that it will break on the newer versions, and provide a command line utility to bring an old file up to the new version in the most sensible and stable way. i.e. ptypy_format_update foo.ptyd would, where possible take foo.ptyd and make it compatible.

Might be a bit hard to maintain this, but you could guarantee it only for a transitionary period between releases. e.g: one major release.

stefanoschalkidis:
You're welcome Björn and thanks for adding the annotation!

I will have to do some research on this first but for now I would recommend to announce if any changes are to be applied to the .ptyd structure.

bjoernenders (approved):

PR #101, opened 2017-09-14
Unittests for data.py

stefanoschalkidis:
These unittests currently have a coverage of only 62%, mainly because file creation and loading would require a dummy file for comparison purposes. Do you think it would be ok to have such a file in resources maybe, or would that clock up the repo over time?

Also, for testing DataSource I would need a dummy dictionary of scans. If anyone has a structure they use regularly I’d happily include it instead of creating a new one.

### DIFF
--- a/ptypy/test/core_tests/test_classes.py
+++ b/ptypy/test/core_tests/test_classes.py
@@ -1668,8 +1668,7 @@ class TestPOD(unittest.TestCase):
              'obj': None,
              'exit': None,
              'diff': None,
-             'mask': None, }
-            ,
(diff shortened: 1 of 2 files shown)


bjoernenders:
Hi Stefanos,

`DataSource` will be scrapped in future, there is no point in developing another test.

Also please make this test relevant in that it uses at least a few frames of 16x16 pixels (or larger) and setup the areas for np.allclose comparison at the top of the script. 

Thanks,
Bjoern

pierrethibault:
Obsolete.

PR #113, opened 2017-11-06
Merging 3d and Bragg now that the models have been generalized

alexbjorling:
~~Placeholder PR (don't merge!).~~ Same changes as PR #73 but now targeting release_candidate. @pierrethibault see changes to `View` etc.


alexbjorling:
Compared View creation timing for commit 1fbfe4e with 14d882d (so this is before Bjoern fixed the caching issue). The 2d/3d switches don't add any significant delay that I can see.
figure_1

PR #122, opened 2017-11-13
Automatic citation reminders at end of reconstruction

alexbjorling:
I think it would be nice to put citations directly in the code (engines, non-trivial scan models, etc.) and to have a summary of papers that should be cited at the end of the run.

I've seen this in other packages (like gromacs) and I think it might motivate people to contribute.

### DIFF
--- a/ptypy/core/ptycho.py
+++ b/ptypy/core/ptycho.py
@@ -185,6 +185,9 @@ class Ptycho(Base):
         # Early boot strapping
         self._configure()
 
+        # Keep a bibliography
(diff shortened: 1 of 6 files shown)


alexbjorling:
Given last night's discussion, should this PR be directed at `release_candidate` instead?

pierrethibault:
I did a few tweaks, and we will merge this as soon as 0.3 is released.

pierrethibault:
Squashed and patched onto release_candidate

Issue #67, opened 2016-11-03
Unused files


aaron-parsons:
Building on what @stefanoschalkidis noticed in #62, there seem to be a number of deprecated/ unused modules/functions present at the moment. It's very clear from the test coverage stats.

ptypy/core/data_old.py -- last commit 2 years ago
ptypy/core/xy_old.py -- last commit 2 years ago
ptypy/core/xy.py raster_scan_legacy, round_scan_legacy, round_scan_roi_legacy
ptypy/engines/ML_old.py -- June 2016
ptypy/experiment/I13DLS_legacy.py -- Does anyone from I13 still use this?
ptypy/simulations/ptysim.py::simulate_basic_with_pods - deprecated bjoern 24th February 2015
ptypy/core/pattern.py # same as xy.py? Stefanos ticket #62

Is there a reason any of this is still there? I would propose that at this milestone (30th November), we make an annotated tag in master to note the point, then delete anything that is deprecated. Anything that remains should be tested. If someone screams that something is broken we can always restore it.

As far as I can tell, and please correct me, everything in this list except for ML_old.py is fine to go. ML_old I know @stefanoschalkidis was working on recently. Is it done, or shall we keep it a while?

aaron-parsons:
reference PR #134

aaron-parsons:
This is now done.

PR #134, opened 2018-01-02
Unused files #67

aaron-parsons:
This PR looks to complete the suggestions in #67 . 

I hope that's ok. Let me know if I missed something being used, but tests seem to run fine to me, and this was all unreferenced anyway.


alexbjorling:
Note that release_candidate has diverged quite a bit from master and is likely to replace it. The merge will be a nightmare and probably amount to reimplementing master's updates more or less manually.

aaron-parsons:
Aha. In which case I'll close this and get myself up to date with the release_candidate branch instead.

PR #145, opened 2018-02-09
Progress on CUDA stubs

aaron-parsons:
Hi @pierrethibault @bjoernenders @alexbjorling,

Here's a placeholder so you can see what I'm up to on the gpu_cuda branch and add some comments.

Things I am aware of:
1/ The vectorise method needs to go into scan model.

2/ I seem to have some really weird error in templates/minimal_numpy_test.py which uses DMnpy. I currently have no idea why, but I'm now off for a week so will have to look when I come back. It almost gives a moonflower scan using a hybrid of the pod method and direct numpy calls, which is how I currently see the result being able to be used.

@bjoernenders as you can see, the general idea of this is basically the same as your ocl implementation so not much has changed there. The contractor has some good thoughts, which I guess Mark can tell you about on Monday.

Thanks,
Aaron


aaron-parsons:
@darrenbatey

aaron-parsons:
Closing this in favour of #164

PR #164, opened 2018-08-01
Accelerate package

aaron-parsons:
@bjoernenders I've squashed the gpu_cuda branch, and restructured to have an accelerate package.

This is the basis for the next release. 
Next moves are:
1. Change setup.py to make accelerate.cuda an optional dependency.
2. Expose some of the C++ methods so we can talk about API.
3. Change accelerate.array_based to be cython.


bjoernenders:
Outdated.

Issue #192, opened 2018-11-23
Floating intensities broken in ML engine


aaron-parsons:
if self.p.floating_intensities:
                diff_view.float_intens_coeff = ((w * Imodel * I).sum()
>                                               / (w * Imodel**2).sum())
E               AttributeError: 'View' object has no attribute 'float_intens_coeff'

This has come about I think due to moving to slotted classes for the views. Previously we could just attach things to the views at will, but now the API is firmed up we have to make another way to do this. I've made a branch: broken_ML_float_intensities to fix this, which has a broken test.

Issue #191, opened 2018-11-12
Sample and Illumination None behaviour


pierrethibault:
Sub-parameters for `illumination` and `sample` are announced as `None` by default in the docstrings but the actual defaults are not `None`. As a consequence it is currently essential to pass explicitly `illumination.diversity=None`, `sample.process=None`, etc. This is not the expected behaviour.

aaron-parsons:
Yup. It turns out that #203 was actually this. because process wasn't set to None explicitly (even though the default was None), it tried to process the stxm data and screwed up the result:-)
I'm off tomorrow, but might take a look this evening.

At least we can carry on.

aaron-parsons:
Ok. I think that the neatest way to do this is in Param.update. 

Not something I want to touch in hurry so will look again on Monday.

bjoernenders:
This needs to be reviewed if that behavior is still there.

PR #193, opened 2018-11-23
fixes ml float intensities

aaron-parsons:
This is related to #192.

@bjoernenders @pierrethibault What do you think? Are you happy to add this field here? I'm not sure if it's a good use of the API going forward, but is the simplest fix.


aaron-parsons:
Ignore this. It’s a terrible idea. I’ll look at solving this with a dictionary in the engine next week.

aaron-parsons:
I think this is ready to merge, but I need to fix the problem with master first #191

aaron-parsons:
This passes and can be merged @bjoernenders

bjoernenders (approved):
Ok I think this one is fine.

Issue #196, opened 2019-01-29
positions in ptyr output


alexbjorling:
I'm making a new reconstruction viewer for the beamline. It's sometimes nice to overlay the scanning positions on the reconstruction. Any good reason that the positions are not saved to the ptyr file?

aaron-parsons:
I don't think so. It's something I also want at some point. I guess we want the original positions + the final positions of the reconstruction if we have done position correction.

________________________________

bjoernenders:
I don't object in having another saving type that saves the positions. But if your request is to just overlay the positions you can use the coordinate grid provided by the saved object storage and display the positions saved in the .ptyd file. That's how I usually do it. Works fine.

alexbjorling:
No, finding the ptyd file is too complicated (for my users at least).

PR #197, opened 2019-01-30
Saving positions in ptyr output

alexbjorling:
Could it be as simple as this? Issue #196.

### DIFF
--- a/ptypy/core/ptycho.py
+++ b/ptypy/core/ptycho.py
@@ -794,7 +794,7 @@ class Ptycho(Base):
             Type of saving, one of:
 
                 - *'minimal'*, only initial parameters, probe and object
-                  storages and runtime information is saved.
+                  storages, positions and runtime information is saved.
                 - *'full_flat'*, (almost) complete environment
 
         """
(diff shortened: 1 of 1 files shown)


bjoernenders:
Hi Alex,
I see the utility in saving the positions, but I personally would rather have a new saving type instead if enhancing the minimal type, call it "positions" or whatever.  The call to .views is unfortunately resource intensive and does not scale (currently) to large data sets. Having an alternative saving option instead of overloading the present one would isolate performance issues if they occur.
Bjoern

alexbjorling:
Good point. I'll add a new type. But can one choose the saving type through the input parameters?

bjoernenders:
Apparently no. You spotted a design flaw. Since a minor change is not supposed to change the input arguments, we cannot do anything about that on the master branch. So my recommendation is to add an extra kw argument for now and pass it on to the save_run call.

aaron-parsons:
Whats the state with this PR?

alexbjorling:
Not sure, am overwhelmed at the moment. I guess I didn't understand the last comment and left it...

bjoernenders:
Merged in with 8af5df7

PR #194, opened 2019-01-16
Update code for parsing AMO/LCLS experiments

daurer:
This is the code I am currently using to parse AMO/LCLS data, would be great to have it merged into the master branch.

Thanks,
Benedikt

PR #198, opened 2019-03-07
Resample

daurer:
Introducing the parameters "resample" and "pad" which allows the shape of the object/probe model to be different from the raw diffraction data. 

The "resample" feature has so far been implemented in DM and ML, both tested on real data. 

The "pad" feature has not been tested.


bjoernenders (commented):
Hi Benedikt,
thanks for the PR. I understand the need to resample, though I'm not very fond of the if switches in DM and ML. Is there any reason, that this is not part of the Geometry, but handled outside of bw() and fw()? Also the changes to manager defy a bit the purpose of the modularity that I thought we introduced with having mutliple scan Models. I'm trying to understand why code has been inserted into Vanilla and Full and there is not a Resize scan model instead? Thinking of backward compatibility we shouldn't sneak in parameters, but instead facilitate bring up parallel structures. If we allow this to become mainstream there are suddenly new parameters for everyone.
Best,
Bjoern

ptypy/engines/ML.py
```diff
@@ -468,7 +471,10 @@ def new_grad(self):
             for name, pod in diff_view.pods.iteritems():
                 if not pod.active:
                     continue
-                xi = pod.bw(w * DI * f[name])
+                if pod.model.resample is not None:
+                    xi = pod.bw(u.zoom(w*DI, pod.model.resample, order=0) * f[name])
```
aaron-parsons: I agree with @bjoernenders, can't we make this in the geometry.fw /bw?

ptypy/engines/utils.py
```diff
@@ -122,7 +122,10 @@ def basic_fourier_update(diff_view, pbound=None, alpha=1., LL_error=True):
     if LL_error is True:
         LL = np.zeros_like(diff_view.data)
         for name, pod in diff_view.pods.iteritems():
-            LL += u.abs2(pod.fw(pod.probe * pod.object))
+            if pod.model.resample is not None:
```
aaron-parsons: Adding resample in the geometry class would mean that this disappears no?

ptypy/engines/utils.py
```diff
@@ -134,9 +137,11 @@ def basic_fourier_update(diff_view, pbound=None, alpha=1., LL_error=True):
             continue
         f[name] = pod.fw((1 + alpha) * pod.probe * pod.object
                          - alpha * pod.exit)
-
-        af2 += u.abs2(f[name])
-
+        if pod.model.resample is not None:
+            af2 += u.rebin_2d(u.abs2(f[name]), pod.model.resample)[0]
```
aaron-parsons: If .fw /.bw were changed then this switch is not needed.

ptypy/engines/utils.py
```diff
@@ -151,7 +156,10 @@ def basic_fourier_update(diff_view, pbound=None, alpha=1., LL_error=True):
         for name, pod in diff_view.pods.iteritems():
             if not pod.active:
                 continue
-            df = pod.bw(fm * f[name]) - pod.probe * pod.object
```
aaron-parsons, line 155: if pod.bw included the sampling then we would not need the following switches/

ptypy/experiment/AMO_LCLS.py
```diff
@@ -9,104 +9,72 @@
...
-
 logger = u.verbose.logger
 
 # Parameters for the h5 file saved by hummingbird
 H5_PATHS = u.Param()
 H5_PATHS.frame_pattern = 'data/photons'
 
-
 @register()
-class AMOScan(core.data.PtyScan):
```
aaron-parsons: This looks fine, but maybe take a look at the Hdf5Loader class that I wrote? Everyone at Diamond now uses this class and I'd be happy if you wanted to extend it to cover your uses.

aaron-parsons:
Hi @daurer @bjoernenders,

I added a few inline comments that agree with Bjoerns geometry comments.

The comments about model and engine code additions I will post on #200 

Hope that's ok.

Aaron

aaron-parsons:
@daurer Could you also add some form off test to the /tests folder? This will protect it against future refactor. It doesn't need to be a full blown unit test at this stage, but a moonflower scan that does it would be ample.

aaron-parsons:
Also, please rebase onto master to make the tests pass.

aaron-parsons:
Ok I rebased this, waiting for the tests to come through.

aaron-parsons:
I didn’t do that properly. My fault for rushing. Will look at it tonight or tomorrow morning.

aaron-parsons:
The next stage on this is to split it into 2 PR's to separate from the AMO stuff and the upsampling. The AMO can just be merged, but the upsampling needs some changes. Happy to do these, and will either happen this week or early next.

daurer:
Hi Bjorn and Aaron,
Thanks for having a look and sorry for the delayed response (I am on vacation at the moment). I agree with your points above and will have a look next week on how to improve the upsampling implementation. 

As for the next steps, I also agree that It should be easy to merge the AMO stuff (#194) into master and then fix/rebase the upsampling PR.

I can do all of that next week, but feel free to go ahead before if you have time.

daurer:
I have now squashed everything into a single commit - introducing changes to scanModels and engine in order to allow for resampling and padding as new features. I have also rebased this commit on top of the current master (with current tests passing).

Next steps are 
1) write a simple test case for resampling ☑
2) write a simple test case for padding @pierrethibault could you take care of this? I don't really know the use case for this feature
3) make necessary changes regarding #200 

About hiding the resample switches of the engines inside the geometry, I have included some comments in response to Aaron's comments. I agree with you that it would be nice to avoid those switches, but I currently see no way around without making things even more complicated. 

Another solution might be to make the default of the ```pod.resample``` to be 1 (currently it is None) and always apply the resampling operations. This would of course create some overhead for the normal user, so not sure if you are okay with that. Just wanted to mention this is another option.

ptypy/engines/utils.py
daurer: The resampling always happens on the intensities. The argument and return value of both bw/fw are complex valued. In order to hide the resampling in the geometry class, we would need a new method in the geometry class that returns ```RESAMPLE(u.abs2(self.fw(x)))``` where ```RESAMPLE``` would be new resample switch.

ptypy/engines/utils.py
daurer: Yes, if the geometry class gets a new function that handles resampling on the intensities and returns those intensities, this switch could disappear.

ptypy/engines/utils.py
daurer: Here things are a bit more tricky. In the resampling case, its again only the amplitudes that are resampled, while the phases (part of ```f[name]``` remain untouched. In order to elegantly hide all the resampling inside the geometry class, we would need to allow ```fw``` to accept two separate arguments (```fm and f[name]```). I think this would make things a bit complicated (more than it currently is)

daurer:
Here is the output of the MF template using "resample=2" and DM  (```minimal_prep_and_run_resample_DM.py```):

MF_resampled_DM

compared to normal DM (```minimal_prep_and_run.py```)

MF_original_DM

daurer:
With ML (```minimal_prep_and_run_resample_ML.py```) there seems to be some issue with the current resampled implementation:

MF_resampled_ML

ptypy/engines/ML.py
daurer: The problem with moving the resampling to geometry.fw/bw is that the resampling only happens on ```w*DI``` while ```f[name]``` remains unchanged. So it would be necessary to pass those arrays as separate arguments to geometry.bw. I am not sure if that is a good solution.

bjoernenders:
@pierrethibault @daurer  This may be a dumb question, but why is the data not resampled and padded when placed into the storages? At the cosmic beamline that data is interpolated and padded all the time before the algorithm sees it. Do you want to preserve noise characteristics? And if that is the case, can we not determine the statistical weights first and then interpolate them along with the data?  Or are you using resample to shrink / grow the state vector dynamically to accelerate algorithmic conversion?

pierrethibault:
The main point of this feature is to deal with data that is undersampled. In this case interpolation would not be appropriate, and instead it is the diffraction-plane constraint that must be adapted. The most general case would require also a detector convolution kernel, but for now it is assumed that it is a box (the detector pixel). So for instance with a resampling of 2, the DM Fourier projector modifies minimally the FFT of the exit wave such that the summed intensity over 4 pixels is equal to the measured detector pixel.

@daurer sorry I read only now many of the other comments on this thread. We would need to look at what makes the fmag error increase (it seems to be an artefact because the other two errors are valid). Regarding ML, I think this is more of a convergence issue. Maybe the initial probe is too far off?

bjoernenders:
@pierrethibault I see. Attaching a convolution kernel to the propagator can be done simply be multiplying a sinc on the input and comes at almost no computational cost ( as compared to the zoom operation) . You can adapt the sinc to whatever the zoom level is for you data rescaling. Would a data interpolation then be ok (maybe plus a one time convolution)?

daurer:
@bjoernenders @pierrethibault @aaron-parsons 

I would like to bring this "upsample" feature into master. It has become a very valuable tool when dealing with data that is poorly sampled, usually because of long tails in the beam (e.g. KB optics).

I have read again through all the feedback on this PR and also discussed in detail with Aaron. I currently see no other possible implementation for DM and ML which would achieve the same outcome. But I have tried to minimize the number of switches in the engines and instead decided to use u.zoom and u.rebin_2d as wrappers. In the case of zoom, the function now returns immediately if there is nothing to be zoomed (the standard case) and rebin_2d just reshapes to the same shape. 

For the changes in the scan model, please check that the current changes are compatible with other feature upgrades that are in the pipeline (e.g. Block model, ...) 

Also, to my understanding, this PR is fully backwards compatible. For the normal user, resample will be equal to 1 and everything works as usual without any drop in performance.

Please let me know what you think!

Benedikt

daurer:
UPDATE: I have decide to make proper upsample/downsample wrappers and put them into the geometry class. I also made those wrappers available through the pods. I think this provides a more clean and readable implementation. It also makes it easy to later expand the upsample function to use different deconvolution kernels (e.g. sinc). 

I have also found a small bug, making sure now that total power is preserved when upsampling.

ptypy/engines/posref.py
```diff
@@ -103,7 +103,7 @@ def estimate_fourier_metric(self, di_view, obj):
         '''
         af2 = np.zeros_like(di_view.data)
         for name, pod in di_view.pods.items():
-            af2 += u.abs2(pod.fw(pod.probe*obj))
+            af2 += pod.downsampe(u.abs2(pod.fw(pod.probe*obj)))
```
aaron-parsons: typo @daurer

ptypy/test/template_tests/ptypy_i13_AuStar_nearfield_9p7keV_test.py
```diff
@@ -99,4 +100,4 @@ def test_i13(self):
         P = Ptycho(p,level=5)
 
 if __name__ == '__main__':
-    unittest.main()
\ No newline at end of file
+    unittest.main()
```
aaron-parsons, line 103: @daurer 
newline

aaron-parsons:
Hi @bjoernenders @pierrethibault ,

Just reviewed this with @daurer. There are a couple of typos which benedikt is fixing now, but otherwise I think this looks like the cleanest implementation it will be within the current framework.

I plan to merge this on Monday evening in the UK unless I hear anything from you guys.

Aaron

ptypy/test/template_tests/ptypy_i13_AuStar_nearfield_9p7keV_test.py
daurer: there is actually a newline

Issue #190, opened 2018-11-12
Overriding PtyScan parameters


pierrethibault:
Sometimes a ptyd file's metadata (`distance`, `center`, ...) has to be tweaked. It would be good to have the possibility to override these parameters directly in the input script. For instance:
```python
p.scans.my_scan = u.Param()
p.scans.my_scan.name = 'Full'
p.scans.my_scan.data = u.Param()
p.scans.my_scan.data.name = 'PtydScan'
p.scans.my_scan.data.source = 'my_prepared_data.ptyd'
p.scans.my_scan.data.distance = 7.5 # instead of 6.5 in meta
```
To keep with the current way things are done, having `source = 'file'` still would not allow additional tweaking:
```python
p.scans.my_scan = u.Param()
p.scans.my_scan.name = 'Full'
p.scans.my_scan.data = u.Param()
p.scans.my_scan.data.name = 'PtydScan'
p.scans.my_scan.data.source = 'file'
p.scans.my_scan.data.dfile = 'my_prepared_data.ptyd'
p.scans.my_scan.data.distance = 7.5 # This field will be ignored
```

pierrethibault:
Implemented.

pierrethibault:
Some tests are now failing because of these lines:
https://github.com/ptycho/ptypy/blob/cab72ed2fb3359d35e44aa42f2a8f23285e82cd0/ptypy/test/ptyscan_tests/on_the_fly_ptyd_test.py#L66-L71
Passing `PtydScan.DEFAULT` to the `PtydScan` constructor including a `source` will override all of `meta`.
I would say that this is not a problem and that the test has to be modified, but happy to reconsider. My main problem is that I don't know how to do this. @aaron-parsons ?

aaron-parsons:
Sure. I'll do this.

________________________________

daurer:
It looks like this has been fixed. Closing the issue.

PR #199, opened 2019-03-11
Fix build

aaron-parsons:
#190 

Hi, 

Doesn't this fix the build?
I thought the point with the new parameter structure was that you didn't need to do:
data = PtydScan.DEFAULT.copy()

You just define a Param with only the things you want, and the default gets taken.

Working with copies of dictionaries is always risky.

@bjoernenders @pierrethibault can you take a look and tell me if I missed something?

I need to tidy up this test a bit more anyway, but I can see what @bjoernenders wants from it.

Aaron


bjoernenders:
Can't see anything wrong with it, so I'm going ahead and approve this.

PR #201, opened 2019-03-18
Orthogonal Probe Relaxation (OPR)

daurer:
To ease a future merge of the OPR feature, I have decided to squash all previous commits into a single one and fixed all conflicts along the way. This obviously needs a careful review. (Not sure if squashing was a good idea, just thought having a clean history might help in this case...)

The biggest change now is the introduction of new engines DMOPR and MLOPR, including some utility functions.

But there are also code changes related to FERMI. They have been moved to a new branch "FERMI" now.

Next steps:
- Include tests for DMOPR and MLOPR
- check DMOPR and MLOPR engines against changes that have been made in DM and ML


aaron-parsons:
Hi Benedict,

It's better to rebase on to master and squash the commits right at the end.

That way we can review easily it along the way, but at the end the feature goes in as a single commit.

In order to do this, we really need to not have the fermi and i13_ffp changes in there. 

I know this is a pain, and I can help if you get stuck, I'm also starting to look into this now. The goal is to make MLOPR and DMOPR engines disappear and just be DM. We start work on a OPR scan model here at Diamond, so I hope to finish this.

Aaron

aaron-parsons:
Also, you have to remember (as I just did when I pulled this branch), when you rebase, you do a force push. This makes the history now different on this branch compared to what everyone else has.

So when I pulled this on to my existing local OPR branch, I get all sorts of horrible conflicts. Instead I have to delete the local branch and pull the remote as fresh. 

So this is why the rebase and squash should be the last thing before a merge is what I mean to say...

aaron-parsons:
When I try to run the template now I get:
AttributeError: 'DMOPR' object has no attribute 'pbound'
Am I missing something?

aaron-parsons:
I'll create a test from this tempalte now.

daurer:
Hi Aaron, 

thanks for your comments. Sorry for the mess. I realise that squashing at this stage was a bad idea (as it rewrites history). 

The FERMI stuff is there because @simonesala was developing the OPR code while working on his FERMI dataset. I have separated those changes now into a new FERMI branch.

daurer:
I justed tested the template ``` simulation_test_OPR.py ``` and it runs fine on my machine.

aaron-parsons:
Great thanks! 

it's completely fine :-) I can see how the history becomes messy very easily at the moment so don't worry. You're completely right though. What we want eventually though is to make it so that it's a single commit that goes into ptypy adding the OPR functionality. 

That way it makes it easy for @bjoernenders to see what's going on.

aaron-parsons:
> I justed tested the template `simulation_test_OPR.py` and it runs fine on my machine.

Works for me now too. Don't know what I did though :-)

aaron-parsons:
OK. I did the simplest thing I could think of to get the new scanmodel running with DM. It is not pretty at all, but basically copies and pastes the part from engine.initialize into a new scan model.

Once I get this working, I want to change the implementation so that the OPR modes are storages. This will require quite a bit of work to get it to work with the Full scanmodel.

I am currently ditching other new parameters like this IP probe thing that can come in as other constraints in their own PR's. They don't seem to require scanmodel, and thats what I am focussing on here.

Comments/criticisms welcome :-), I have a meeting and then will be continuing.

aaron-parsons:
For difference map, both now run through the same engine, with only a single switch needed in the probe update. I have found that I get different results with the DM probe update that came from the OPR branch, compared to the standard issue one, and in fact I think I trust the one from the OPR branch more since it gives transmission values between 0 and 1. I'd really like someone else to check this other than me. At a glance, the mpi performance seems to be similar with the same, Full model implementation.
Otherwise, the to-do list is:
1.  Get the MLOPR implementation running in a similar way, using a scan model instead of a different engine.

2. Refactor the scan model to be software and not a hack.The reason to focus on this is due to the GPU extension in future. I would prefer to make it use storages rather than the view mapping we currently have.

3. Benchmark the engines with the new model.

4. Tidy up and improve tests.

5. Finish PR.

aaron-parsons:
We also need to make the OPR model compatible with the regular, full model (even though a performance impact might be noticed). I reckon we can do this with a switch to link the pods.

daurer:
> I have found that I get different results with the DM probe update that came from the OPR branch, compared to the standard issue one, and in fact I think I trust the one from the OPR branch more since it gives transmission values between 0 and 1. I'd really like someone else to check this other than me.

I have tested this (without MPI) and I got the following observation:
1) Running ```minimal_prep_and_run.py``` with the "Vanilla" scan model on **master** gives transmission values between 0 and 236, and the probe has a power of ~1e3. 
2) Running ```minimal_prep_and_run.py``` with the "Full" scan model on **master** gives transmission values between 0 and 1, and the probe has a power close to 1e8.
3) Running ```minimal_prep_and_run.py``` with the "Vanilla" scan model on **this branch** gives transmission values between 0 and 0.027, and the probe has a power of 1.3e9. 
4) Running ```minimal_prep_and_run.py``` with the "Full" scan model on **this branch** gives transmission values between 0 and 0.025 and the probe has a power close to 1.6e9.

So I suspect that there is something wrong with the "Vanilla" scan model (on master) and that there is something weird with the probe normalization in the probe_update on this branch.

pierrethibault:
I am not sure there is anything wrong with any of these. The relative scale of probe vs object is an unconstrained degree of freedom that depends on initial conditions and on the engine dynamics. We cannot rely on this to judge the success of a reconstruction. 

We could impose a scale, for instance by forcing the object transmission to always be between 0 and 1, or the probe's square norm to be equal to the maximum photon count in the diffraction data. This would have the benefit of being reproducible, but it would be physically wrong in many instances.

aaron-parsons:
@bjoernenders  could you please take a look at the two DM implementations and let us know your thoughts? There's some pressure now to get this out.

Also I want to make some model converters that can flip between the Full and OPR models. @daurer @pierrethibault Do you think the following behaviours are reasonable?:

1/ Full--> OPR model conversion, just copies the probe array into the new buffers that are needed.
2/ OPR-->Full model conversion. This is more tricky. I am thinking that updating the probe positions from the centre of mass of the OPR mode gives us the position correction, and that updating the object weights (or even directly scaling the .diffs) from the summed intensity of these modes would be reasonable.

What do you think?

daurer:
@aaron-parsons sorry for the delayed response. I generally think your suggestions for converting between the Full and the OPR model are reasonable. I am not sure if I can follow the suggested updates for the OPR->Full case. I guess the idea is to keep the average probe from the OPR model and update the scan positions to keep some information about the variation in the probes, is that correct?

daurer:
@bjoernenders @pierrethibault @aaron-parsons 
I have picked up the OPR project again. And after some days of struggling I now have something that replicates the old successful DMOPR engine, but is merged into regular DM and using Aarons OPR scan model at the cost of being an ugly hack inside the probe update. So my plan for the next days:

- Move all the engine-specific DM code back to a new DMOPR engine, which inherits from the current DM, but just modifies the probe_update function. This keeps the regular DM untouched and only modifies relevant code (instead of copying the whole DM as the case with the previous DMOPR engine)
- Make DMOPR more robust (I can see some convergence issues that I would like to improve)
- Update DMOPR test function and template
- Switch over to MLOPR and implement things with a similar strategy: OPR scan model and lightwight MLOPR engine hooking into regular ML.
- Update MLOPR test function and template

If you have opposing comments, please let me know. Otherwise, I will just continue on this path.

daurer:
I think the implementation is now in a good state. It makes minimal changes to the core and utilizes the "hook" concept for introducing new independent probe features. 
@bjoernenders  @aaron-parsons  please have a look and let me know what you think?

daurer:
My next goal is to work on the OPR algorithm itself. I noticed that I get different convergence results depending on how many CPU cores I am using. I have a feeling that this has something to do with the "round-robin" implementation in ```reduce_dimension```. @pierrethibault do you agree?

Here are some plots after running the template: ```simulation_test_OPR_scanmodel.py```:
image
image
image
image
image

Legend: The white crosses in the probe overview are the expected center-of-mass positions and the red crosses are the per-probe calculated center-of-mass positions.

I am bit worried that some of the probes are clearly not converged, so my plan is to investigate why this happens to them (its not always the same ones, but there are always some).

pierrethibault:
Hi Benedikt. I don't know what causes this. The round robin part is only meant to compute matrix elements as efficiently as possible across the nodes, so it should not have any impact. If it does, it's a bug. Are we sure there aren't some random number generation going on? That also throws MPI off. Thanks for looking into this. Let me know if you need more input from my side.

daurer:
Hi Pierre. I have made a modification in DMOPR on how the change is calculated in the probe_update and seems to have improved the MPI-related effects. I am not sure if this is completely fixed now, but its really hard for me to tell. There is also a certain randomness with each reconstruction, so debugging this is a bit tricky. 

Note that the probe_update of DMOPR is different from regular DM (mostly because probes are independent and any MPI reduction across nodes is taken care of differently), currently there is also no cfact and therefore no probe_inertia applied. 

Anyway, DMOPR seems to work very well now with the simulation (especially once I raised the fourier_relax_factor):
image
image
image

With the MLOPR I am a bit confused that the error jumps at the start.

pierrethibault:
Thanks for the update. Looks good. RE ML error jump, it's something I observed often, and I am pretty sure it is related to the regularisation term in the cost function. That's why I like to use some smoothing in DM, just to keep the edges smooth.

I will try and look at your reimplementation on probe update and will comment when I can.

bjoernenders (approved):
All fine with me now! Go ahead with merging

Issue #146, opened 2018-02-28
Position Refinement with multiple processes


wfeschen:
Hi,

I want to implement a position refinement algorithm [1] to the existing DM-engine. My implementation works fine when I run Ptypy in a single process.
`mpiexec -n 1 python scripy.py`
But when I run the same script on multiple processes e.g.
`mpiexec -n 2 python script.py`
python freezes at the first iteration where I update the positions. 

In the position refinement algorithm I generate random shifts for each position and calculate the Fourier Error for the new positions. If the Fourier Error decreased, I keep the shifted position otherwise I will keep the old position. I did this by editing the engine_iterate method in the DM class. 

```
    def engine_iterate(self, num=1):
        """
        Compute `num` iterations.
        """
        to = 0.
        tf = 0.

        for it in range(num):
            t1 = time.time()

            # Fourier update
            error_dct = self.fourier_update()

            t2 = time.time()
            tf += t2 - t1

            # Overlap update
            self.overlap_update()

            t3 = time.time()
            to += t3 - t2

            # count up
            self.curiter +=1

        logger.info('Time spent in Fourier update: %.2f' % tf)
        logger.info('Time spent in Overlap update: %.2f' % to)
        error = parallel.gather_dict(error_dct)

        """
        Position refinement
        """
        if self.p.pos_ref_stop > self.curiter >= self.p.pos_ref_start and self.curiter % self.p.pos_ref_cycle == 0 \
                and self.p.pos_ref:
            self.pos_ref()           

        """
        Pos refinement end
        """

        return error
```

In the pos_ref() function, which I call at the end of the engine_iterate function I cycle through all pods and try to refine the positions.

```
    def pos_ref(self):
        pod_names = self.pods.keys()
        pod_names.sort()

        for pod_name in pod_names:
            pod = self.pods[pod_name]
            if pod.active:
                # calculate random positions
                ......
                pod.ob_view.coord = shifted_position
                self.ob.reformat()
                error = self.get_fourier_error(pod_name)
                .....
               
```

I always call the `self.ob.reformat()` method when I change the coordinates which is somehow critical, because without calling this line python will not freeze. Is this even the intended way to update the pods to the new coordinates? Or is there a better way?

I also figured out that python is not freezing when I do not use the `if pod.active:` statement, I am not sure, but I think I need this for mpi? I am thankful for all help.

Best
Wilhelm

 
[1] https://www.sciencedirect.com/science/article/pii/S0304399112001222

[...]

PR #206, opened 2019-04-17
Adds the axis grids to the saved files

aaron-parsons:
This just adds the axes to the saved files. This is needed for any downstream analysis of the data.

Issue #200, opened 2019-03-11
Ptypy extension philosophy


pierrethibault:
How to implement a new feature is a question that comes up quite often in discussions with @aaron-parsons lately.

Subclassing ScanModel and Engine seemed the cleanest by design, but there are a few important hurdles. First, many of ptypy's methods are very long, so a subclass is likely to reproduce a large part of the parent's code just to introduce a few lines somewhere in the middle. This code duplication makes maintenance difficult. The solution would be to break up long methods in smaller chunks. I'd be happy to see that, but I wonder if there is a cost in code readability and performance.

The other extreme is to pack existing ScanModels and Engines with more features. This way there is much less code duplication, but we end up with all sorts of switches everywhere, and an explosion of parameters that most users would not need.

This issue is showing up now with Benedikt's (@daurer) pull request, on which I also worked (#198), which is implemented according to the latter scenario. I understand Bjoern's (@bjoernenders) suggestion of creating instead a new ScanModel, and his reticence with the switches deep in DM and ML, but the alterative, as it stands, is to duplicate a lot of code, and possibly to add complexity. For instance, how to deal with combining features, such as resampling and (upcoming) OPR or multislice? Yet another Engine subclass that supports both?

I'd like to hear your thoughts on this.

aaron-parsons:
My thoughts are that there are two separate things. The new ScanModel stuff
is to get rid of anything that would have previously been hacked into engine.prepare.

I think there is a future refactor in ScanModel to make the methods shorter,
but that currently we don't have enough information yet to guide this refactor. This is
why accepting PR's like Benedikts is important. Since the ScanModel has the name Full,
it should either be renamed, or be allowed to contain everything where possible.

Where this isn't possible we should make a new model which explicitly highlights
the differences via inheritance. This might involve refactoring Full along the way.

I also have the same cringe with the switches in the engines. This is also a
future refactor in the making. Alex had to do this too with his Bragg style engines, and so points out that something is amiss.

I am slowly in the process of making the multislice and opr models, and will
push these too.

I would suggest that we accept these PR's so long as they have some form of tests
to guide the future refactor.

From my point of view, I would then accept this PR pending a future core refactor when we understand the limitations.

alexbjorling:
I mean I have no opinion other than that I've noticed the problem and have wanted to chop up long methods a few times, like here
https://github.com/ptycho/ptypy/pull/113/files#diff-6256976fd8a26060572b2cb350cc0eb7

[...]

Issue #21, opened 2016-08-03
Python 3 compatibility


pierrethibault:
Has to be done at some point.

stefanoschalkidis:
Hi all,

One of the reasons I started working on unittests was to make the code Python 3 compatible. While I first thought it would be good to have all unittests ready before starting the rewrites, I now believe it is better to do them right away because of fresher memory. This is why I would like to discuss and agree on certain Python 3 design choices.

- String ambiguity
There are three ways to approach this:

1. We treat all strings as text data in Python 2 and mark the ones containing binary information. This would retain ambiguity of unmarked strings when run under Python 2.

2. We use unicode_literals to transform all strings in Python 2 to unicode and mark the ones containing binary information. This way we remove the ambiguity.

3. We mark all strings on whether they contain text or binary data.

Here is a small pro & con on unicode_literals:
http://python-future.org/unicode_literals.html

- Rewrites
When rewriting the code I would choose the solution which works better for Python 3, e.g.
`for (key, value) in heights.items():    # inefficient on Py2`
Here is a nice cheat sheet containing a few of such cases:
http://python-future.org/compatible_idioms.html

Let me know what your opinions are on this issue and how you would like the code to evolve.

alexbjorling:
How about just migrating from one release to the next, breaking 2.7 compatibility? It is a bit harsh, but the code will look nicer and be easier to work with. I think people can cope, or do they have ptypy built into other applications?
See also #207

jcesardasilva:
Dear All, 
How is the progress on the port of Ptypy to Python 3? 
I've done a fork and translated all the code as much as I could to Python 3 in terms of syntax. However, it seems that there is a problem with the package utils.descriptor. I thought it was only the ConfigParser problem, since RawConfigParser is deprecated, but not sure. 
If you think it is useful for further progress, I can propose a pullrequest with that version and we could work on top of it. However, I think that maybe someone else already did it. Could we have a branch such as "port2to3" to do it? 
The version I made is not fully working yet. I get the following error message when I try to import the version for Python3 I prepared for ptypy: 
```
---------------------------------------------------------------------------
DuplicateSectionError                     Traceback (most recent call last)
<ipython-input-1-7e5d8b111290> in <module>
----> 1 import ptypy

~/.local/lib/python3.7/site-packages/ptypy/__init__.py in <module>
     73 from . import utils
     74 from . import io
---> 75 from . import experiment
     76 from . import core
     77 from . import simulations

[...]

PR #209, opened 2019-05-14
Position correction

daurer:
This is an implementation of a position correction algorithm as described in:

>  A.M. Maiden, M.J. Humphry, M.C. Sarahan, B. Kraus, J.M. Rodenburg,
>         An annealing algorithm to correct positioning errors in ptychography,
>         Ultramicroscopy, Volume 120, 2012, Pages 64-72

It has originally been implemented by @wfeschen, the full story behind this implementation is described in issue #146 

I have briefly tested the performance of this algorithm with DM (```ptypy/templates/position_correction.py```) but certainly more testing (e.g. ML) and other changes (e.g. to make things compatible with the general ptypy code philosophy) is required.


daurer:
Added lasted updates from @wfeschen, including an evaluation script for the corrected positions.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,281 @@
...
+            dcoords[i + 1, :] = dcoord
+
+            # Check if these are really necessary
+            if di_view.pod.ob_view.dlow[0] < 0:
+                di_view.pod.ob_view.dlow[0] = 0
+            if di_view.pod.ob_view.dlow[1] < 0:
+                di_view.pod.ob_view.dlow[1] = 0
+
+            new_obj = np.zeros(self.shape, dtype=np.complex128)
+            if ob_view_temp.data.shape[0] != self.shape[0] or ob_view_temp.data.shape[1] != self.shape[1]:
```
bjoernenders: maybe use np.allclose() here

ptypy/engines/posref.py
```diff
@@ -0,0 +1,281 @@
...
+                # if the data of the view has the wrong shape, zero-pad the data
```
bjoernenders: should we not one-pad the data

ptypy/engines/posref.py
```diff
@@ -0,0 +1,281 @@
...
+                # new data for calculating the fourier transform
+                # calculate limits of the grid
+                ymin = self.engine.ob.storages.values()[0].grids()[0][0, 0, 0]
```
bjoernenders: I would maybe create reference to the object storage like `ob = di_view.pod.ob_view.storage` at the beginning of the function call ot in this if statement

ptypy/engines/base.py
```diff
@@ -62,6 +63,42 @@ class BaseEngine(object):
     help = Valid probe area as fraction of the probe frame
     doc = Defines a circular area centered on the probe frame, in which the probe is allowed to be nonzero.
 
+    [posref]
+    default = False
+    type = bool
+    help = If True refine scan positions
+
+    [posref_start]
```
bjoernenders: We should better use non-flat parameters here. `[posref]` is a group and  `[posref.active]`and then `[posref.start]`

bjoernenders (commented):
Hi Benedikt.
I made only a few comments. Am I correct in assuming that you added the parameters to BaseEngine to make the refinement available in both DM_pos_corr and ML_pos_corr ? If so I suggest subclassing from BaseEngine to PosRefEngine and using that engine as base for the two above. Otherwise the parameters for position refinement appear for all other engines breaking backward compatibility.

daurer:
Hi Bjoern,
thanks for your comments. ```DM_pos_corr``` and ```ML_pos_corr``` are actually leftovers from the initial implementation by @wfeschen. I made the changes to the BaseEngine to make the position correction work with the normal ```DM``` engine. This is because I wanted to use other new features that I have implemented in DM (e.g. resample). This touches on the questions raised in #200. 

Unfortunately, I currently have no time to work on this PR, but I hope that @aaron-parsons (maybe with help from @wfeschen) can take over from here and make this position correction idea into a proper ptypy extension.

aaron-parsons:
Hi all,thanks for all the work/comments. I'll be following this up on Friday since we need it on a couple of beamlines.

aaron-parsons:
Also, yes I agree with Bjoern with the PosRefineEngine (or similar).That's essentially what the position refine class is doing here, but passing the engine through rather than extending it. It should be pretty simple for me to do this.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,281 @@
...
+        # Shape and pixelsize 
+        self.shape = self.engine.pr.S.values()[0].data[0].shape
+        self.psize = self.engine.ob.S.values()[0].psize[0]
+
+        # Maximum shift
+        start, end = engine.p.posref_start, engine.p.posref_stop
+        self.max_shift_dist_rule = lambda it: self.amplitude * (end - it) / (end - start) + self.psize/2.
+
+        # Save initial positions to file
+        self.save_pos()
```
aaron-parsons: Is there a reason to save the positions here? Now that we dump out the positions at the end anyway #206 I'm thinking not.

templates/position_evaluation.py
```diff
@@ -0,0 +1,195 @@
...
+        plt.xlabel("Position number")
+        plt.ylabel("Position number")
+        plt.colorbar()
+
+        tot_dist_uncorr = np.sqrt((relativ_shifts_uncorrected[:, :, 1] - relativ_shifts_original[:, :, 1])**2 + (relativ_shifts_uncorrected[:, :, 0] - relativ_shifts_original[:, :, 0])**2)/psize*1e-6
+        print("Mean distance error to all other positions corrected: " + str(np.mean(np.mean(tot_dist, axis=0))))
+        print("Mean distance error to all other positions uncorrected: " + str(np.mean(np.mean(tot_dist_uncorr, axis=0))))
+        plt.legend()
+
+    plt.show()
```
aaron-parsons, line 195: Update this script to show it working for the output .ptyr

templates/position_refinement.py
```diff
@@ -0,0 +1,74 @@
...
+a = 0.
+
+coords = []
+for pname, pod in P.pods.iteritems():
+    # Save real position
+    coords.append(np.copy(pod.ob_view.coord))
+    before = pod.ob_view.coord
+    new_coord = before + 3e-7 * np.array([np.sin(a), np.cos(a)])
+    pod.ob_view.coord = new_coord
+
```
aaron-parsons, line 65: We have a similar example for the OPR branch. I wonder if it makes sense just to add it as an option for the moonflower scan.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,281 @@
...
+from .. import utils as u
+from ..utils import parallel
+from ..utils.verbose import log, logger
+import numpy as np
+import os
+import sys
+import time
+import gc
+
+class PositionRefine(object):
```
aaron-parsons, line 17: I think we can split this up a bit. I think PositionRefine should do the calculation, but the iteration through the views should be done in the PositionRefinementEngine class. We can also allocate temp_ob in the engine initialize method then too.

ptypy/engines/base.py
aaron-parsons: I have done this, but #191 is causing issues separate to this PR.

ptypy/engines/posref.py
aaron-parsons: Actually no padding is done here.

ptypy/engines/posref.py
aaron-parsons: Yes. I have now done this.

aaron-parsons:
Not sure why the CI is failing, the test on it's own passes. @bjoernenders @pierrethibault are you happy with this at the moment?

If I fix the CI I am thinking to just merge this. Once my users start working with it, more bugs might become apparent.

One thing I noticed is that I still didn't do #191 which means that it is hard to turn off unless the start and stop are None. Once this one is merged, I'll look at #191

ptypy/engines/base.py
```diff
@@ -275,13 +257,144 @@ def engine_iterate(self, num):
...
+    type = int
+    help = Number of iterations until position refinement starts
+
+    [position_refinement.stop]
+    default = None
+    type = int
+    help = Number of iterations after which positon refinement stops
+    doc = If None, position refinement stops after last iteration
+
+    [position_refinement.cycle]
```
pierrethibault: We have used "interval" for this concept in other places in the code...

ptypy/engines/posref.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Refines the positions by the following algorithm:
+    
+    A.M. Maiden, M.J. Humphry, M.C. Sarahan, B. Kraus, J.M. Rodenburg,
+    An annealing algorithm to correct positioning errors in ptychography,
+    Ultramicroscopy, Volume 120, 2012, Pages 64-72
+    """
+    def __init__(self, position_refinement_parameters, initial_positions, shape, temp_ob):
```
pierrethibault: Definitely need a docstring here.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,190 @@
...
+        # Maximum shift
+        start, end = self.p.start, self.p.stop
+        self.max_shift_dist_rule = lambda it: self.p.amplitude * (end - it) / (end - start) + self.psize/2.
+
+        log(3, "Position refinement initialized")
+
+
+    def fourier_error(self, di_view, obj):
+        """
+        Calculates the fourier error based on a given diffraction and ohject.
```
pierrethibault:
```suggestion
        Calculates the fourier error based on a given diffraction and object.
```

ptypy/engines/base.py
```diff
@@ -275,13 +257,144 @@ def engine_iterate(self, num):
...
+    type = float
+    help = Distance from original position per random shift [m]
+
+    [position_refinement.max_shift]
+    default = 0.002
+    type = float
+    help = Maximum distance from original position [m]
+
+    """
+
```
pierrethibault, line 312:
```suggestion

    def __init__(def __init__(self, ptycho_parent, pars=None):
        """
        Base engine for position refinement.
        """
        super(PositionCorrectionEngine, self).__init__(ptycho_parent, pars)

        self.ptycho.citations.add_article(
            title='An annealing algorithm to correct positioning errors in ptychography',
            author='Maiden et al.',
            journal='Ultramicroscopy',
            volume=120,
            year=2012,
            page=64,
            doi='10.1016/j.ultramic.2012.06.001',
            comment='Position Refinement using annealing algorithm',
        )

```

bjoernenders: We should only add the citation if the position correction has been switched on.

pierrethibault: Good point. Ok, let'sl put it somewhere else.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            new_coordinate = (0, 0)
+
+        # Clean up
+        del ob_view_temp
+        del di_view
+        
+        return new_coordinate
+
+    def reset_access_rule(self):
```
pierrethibault: A docstring here would be good as well.

ptypy/engines/posref.py
pierrethibault: ... actually there is an implicit padding done in line 166, since `new_obj` is initialised as `zeros`. In general I must say I feel that this part of the code is a bit fragile. I wish this type of stuff was folded into the `Storage` or `View` class. Maybe `View.get_data(pad=True)` could return exactly what is computed here on line 166?

Another kind of reasonable approach might be to add the possibility of having extra "safe" padding in a Storage. By padding safely the object storage by the maximum allowed shift, we would know that all views would match, and no test would have to be done at this step. Once all positions are updated we only need to update the safe padding to maintain it. Is the memory cost too high?

pierrethibault: I implemented extra padding (branch extrapadding). I think that could be an elegant way of working with this. @bjoernenders @aaron-parsons @daurer @wfeschen opinion?

bjoernenders: Yes, good idea, I would have put it at the same location. I don't think though that you should need the padding in `__getitem__` and `__setitem__` ... Shouldn't that be automatically resolved by the views taking on higher and lower values for dlow and dhigh respectively after reformat?

bjoernenders: What I want to say is that I would have expected it to just work with the misfit correction in line 658

ptypy/engines/base.py
aaron-parsons: Just added this in the appropriate place.

ptypy/engines/base.py
```diff
@@ -364,8 +364,13 @@ def position_update(self):
             new_coords = {}
 
             # Maximum shift
+
             self.position_refinement.max_shift_dist = self.position_refinement.max_shift_dist_rule(self.curiter)
-            self.position_refinement.reset_access_rule()
+            for sname, storage in self.position_refinement.temp_ob.storages.iteritems():
+                print("Old storage shape is: %s" % str(storage.shape))
+                storage.padding = int(np.round(self.position_refinement.max_shift_dist + 1))
```
aaron-parsons: Debugging here shows that the shape is indeed increased.

ptypy/engines/posref.py
```diff
@@ -124,51 +126,20 @@ def single_pos_ref(self, di_view):
...
-                idx_x_low = 0
-                idx_x_high = self.shape[1]
-                idx_y_low = 0
-                idx_y_high = self.shape[0]
-
-                if new_ymin < ymin:
-                    idx_y_low = self.shape[0] - ob_view_temp.data.shape[0]
-                elif new_ymax > ymax:
-                    idx_y_high = ob_view_temp.data.shape[0]
+                print("I went in here tempshape:%s, self.shape:%s, dx:%s, xy:%s, maxdist:%s" % (ob_view_temp.data.shape, self.shape, delta_x, delta_y, self.max_shift_dist))
```
aaron-parsons: But here the shapes are still wrong. I think that this could be because we pad on the -ve side, do we update the coordinates of the views to be increased by the pad size?

ptypy/engines/base.py
aaron-parsons: Pierre pointed out that the max_distance is in real units, not pixels so this is now fixed.

ptypy/engines/posref.py
```diff
@@ -97,7 +103,7 @@ def update_view_position(self, di_view):
         coord = np.copy(di_view.pod.ob_view.coord)
         
         self.ar.coord = coord
-        self.ar.storageID = self.temp_ob.storages.values()[0].ID
+        self.ar.storageID = self.temp_ob.storages.values()[0].ID # this bit worries me. Shouldn't it get the ID from the di_view?
```
aaron-parsons: Should we be iterating over the object storages?

aaron-parsons: I think the answer is no because it's always that first point of interaction that we care about.

pierrethibault (approved):
Happy to merge.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,185 @@
...
+        delta = np.zeros((self.p.nshifts, 2))               # Coordinate shifts
+        errors = np.zeros(self.p.nshifts)                   # Calculated errors for the shifted positions
+
+        coord = di_view.pod.ob_view.coord.copy()
+        
+        self.ar.coord = coord
+        self.ar.storageID = di_view.pod.ob_view.storageID
+
+        # Create temporary object view that can be shifted without reformatting
+        ob_view_temp = View(self.temp_ob, accessrule=self.ar)
```
bjoernenders, line 155: These views need to be cleaned up. I could envision they will grow to unmanageble numbers.

ptypy/engines/posref.py
```diff
@@ -0,0 +1,185 @@
...
+
+            norm = np.linalg.norm(rand_coord - self.initial_positions[di_view.ID])
+            if norm > self.p.max_shift:
+                # Positions drifted too far, skip this position
+                errors[i] = error_initial + 1.
+                continue
+
+            # Create new view on new position
+            self.ar.coord = rand_coord
+            ob_view_temp = View(self.temp_ob, accessrule=self.ar)
```
bjoernenders: Same here. view_temp is kept forever.

bjoernenders (changes requested):

ptypy/engines/posref.py
```diff
@@ -0,0 +1,185 @@
...
+        # Updated before each iteration by self.update_constraints
+        self.max_shift_dist = None
+        self.temp_ob = None
+        log(3, "Position refinement initialized")
+
+    def fourier_error(self, di_view, obj):
+        """
+        Calculate the Fourier error based on a given diffraction and object.
+
+        :param di_view: View to the diffraction pattern of the given position.
```
bjoernenders: Numpy docstrings everywhere please.

ptypy/engines/posref.py
bjoernenders: Actually let me take care of this. I need to work on the View in faster_views branch anyway

aaron-parsons: I'll answer these in a post at the end

ptypy/engines/posref.py
aaron-parsons: I'll answer these in a post at the end

ptypy/engines/posref.py
aaron-parsons: Done

aaron-parsons:
Right. I think this is ready to go now.

For the question about tidying up the views. Bjoern and I talked about this. What we should really do to encapsulate the behaviour of the .copy() process is copy and then immediately set container.original to be itself, removing the link to the parent object. Any new views are then created only on the new copied instance and not on the parent.

The problem with this is that, from the Container docs:
    _In order to reduce the number of :any:`View` instances, Container copies
    do not hold views and use instead the Views held by the original container_

This means that after setting container.original to itself you would then have to copy all the views to the new container before doing anything else. **This** seems to me to be a bad idea and would definitely impact performance.

Instead what I have implemented is that at the end of each iteration we delete our container copy and create a fresh copy each time. If the views are somehow kept in memory they will be overwritten by the next iteration anyway.

I think that this needs to be looked at more fully in a separate package of work, which is what Bjoern is doing in #212, and what I'll also be looking at as part of our acceleration upgrades next month. As our beamlines start to use this feature I will certainly be able to spend some more time looking at this, but first they have to get the feature.

Hope that makes sense.

aaron-parsons:
Ok now this is an issue for me. When running in MPI mode across 2 nodes with 40 cores each, ON the second iteration of position correction I get the following error.

Iteration #21 of DM :: Time 2.45
Errors :: Fourier 4.35e+03, Photons 5.31e-02, Exit 7.84e-02
Smoothing object, average cfact is 0.00
*** Error in `python': double free or corruption (!prev): 0x000055cea39e72c0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81489)[0x2aea7eb66489]
/dls_sw/apps/openmpi/3.0.0-pty/rhel7/x86_64/lib/libmpi.so.40(ompi_coll_base_allreduce_intra_ring+0x180)[0x2aea89d9a3a0]
/dls_sw/apps/openmpi/3.0.0-pty/rhel7/x86_64/lib/libmpi.so.40(PMPI_Allreduce+0x18b)[0x2aea89d5fd5b]
/dls_sw/apps/ptypy/temporary_environment_rhel7/miniconda/envs/full_dependencies/lib/python2.7/site-packages/mpi4py/MPI.so(+0xdce90)[0x2aea89c57e90]
/dls_sw/apps/ptypy/temporary_environment_rhel7/miniconda/envs/full_dependencies/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x7d14)[0x2aea7e0a4ea4]
/dls_sw/apps/ptypy/temporary_environment_rhel7/miniconda/envs/full_dependencies/bin/../lib/libpython2.7.so.1.0(PyEval_EvalCodeEx+0x7e9)[0x2aea7e0a66c9]
/dls_sw/apps/ptypy/temporary_environment_rhel7/miniconda/envs/full_dependencies/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x6a08)[0x2aea7e0a3b98]
/dls_sw/apps/ptypy/temporary_environment_rhel7/miniconda/envs/full_dependencies/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x7fee)[0x2aea7e0a517e]
/dls_sw/apps/ptypy/temporary_environment_rhel7/miniconda/envs/full_dependencies/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x7fee)[0x2aea7e0a517e]


I have to go now, but any hints are greatly appreciated.

aaron-parsons:
Ok, by moving log statements backwards in the code, I have identified that this happens in the             parallel.allreduce(s.data) in the object update, the iteration after a position refinement has happened.

Adding some more logging and it's now obvious what happens. Across 10 cores on a single node I get:

INSPECT s.data.shape: (1, 446, 442) , s.data.dtype:complex64, rank:9
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:1
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:4
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:8
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:3
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:2
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:5
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:6
INSPECT s.data.shape: (1, 444, 440) , s.data.dtype:complex64, rank:7

So the shape of one of the object storages is wrong. It couldn't just say that I guess....

I think the issue is that I haven't synced the new positions across all cores before doing the ob.reformat. And in the position update I only work with the active diffraction views. This means that each process is getting a different set of co-ordinates to reformat to.
i.e. we set di_view.pod.ob_view.coord, but only for the di_views that are active in each process.

This was mitigated in the original implementation since the positions were stored in one big array that was reduced.

Have to go now again. Tomorrow it shall be fixed.

bjoernenders:
Hey Aaron.
 I uploaded a fix for your mpi problem. Now hopefully, your storages stay synchronized. This fix is also needed for future strong(er) scaling. There might still be issues with this high turnaround of Container copies. I'm also fine with using container copies. Also, `.reformat()` on the copied container does not considers those views you created on it, but only those in the `original`. Just FYI, I hope you don't need that.

But I don't quite understand, why there needs to be an object copy at all and new views, if you can just let the actual `ob_view` wander around. Also, once the overlap breaks, the annealing procedure breaks, cause the algorithm can have the object fit locally perfect with the data, so they cannot really go far, can they? We could also write an additional method for storage, that bounces views back into its data as a safeguard. Stuff to think about, if this procedure needs performance tweaks in the future.

aaron-parsons:
thanks Bjoern. This is still not quite there. When I run this with 10 cores not I get:

  File "/home/clb02321/PycharmProjects/position_correction/build/lib/ptypy/engines/utils.py", line 125, in basic_fourier_update
    LL += u.abs2(pod.fw(pod.probe * pod.object))
ValueError: operands could not be broadcast together with shapes (128,128) (128,0) 
    LL += u.abs2(pod.fw(pod.probe * pod.object))
ValueError: operands could not be broadcast together with shapes (128,128) (128,127) 

So I guess the sizes are wrong... I'll take a look.

aaron-parsons:
This was a problem at my end. We can merge now I think. Thanks everyone.
I'll leave this for @bjoernenders to look at, but otherwise will merge tomorrow morning UK time.

aaron-parsons:
Ok I am going to merge this now. Any feedback is appreciated via the issues. Thanks @wfeschen @daurer @bjoernenders @pierrethibault

PR #214, opened 2019-05-24
Posref fixes

bjoernenders:
Hi @aaron-parsons I created a simpler version of the position refinement that works just as well. @pierrethibault I had to change the initialization with padding, cause in combination with `Storage.copy` it would apply the padding twice. Now padding is only considered on `.reformat()`


aaron-parsons (approved):
Looks good. Much tidier.

PR #215, opened 2019-05-24
Fix sim

aaron-parsons:
HI Bjoern,

I just hit this bug whilst trying to do some sims. It seems like you already applied the fix that I was going to do, but didn't merge.

Was there a reason not to merge?

Aaron


bjoernenders:
Nope, no, reason that I remember. I'll merge this.

Issue #205, opened 2019-04-04
minimal script in tutorial is broken


daurer:
It has been pointed out by @ningustc that the current tutorial is broken. The line 

https://github.com/ptycho/ptypy/blob/2dc39437b453de548ab8a4438f830b1730ee1e8a/tutorial/minimal_script.py#L74

causes a RuntimeError on validation of the parameter tree:

```
RuntimeError: Parameter validation failed:
   - scan
```

pntbttrvbs:
submitted a PR for this. 

A quick fix is to just comment out the problem line.

Basically, when the Ptycho object is initialized, it's trying to validate the parameters. The validate method can't find the default for 'scan', which is set from the docstring by the defaults_tree.parse_doc decorator.

bjoernenders:
Hi, yes the tutorial needs to be changed. Not the class. Thanks for pointing this out.

daurer:
It seems like the tutorial has been fixed.

PR #218, opened 2019-06-24
doc changes to Ptycho class to add 'scan' parameter defaults

pntbttrvbs:
Added default for `scan` parameter to Ptycho docstring to fix tutorial problems (#205 )


bjoernenders:
Thanks for your contribution, but that is not the right fix. The tutorial needs to be adapted, the `scan` entry is deprecated.

pntbttrvbs:
Thanks Bjoern

PR #219, opened 2019-06-26
Hdf5 previewing

aaron-parsons:
Now allows a bounding box to be placed on the input datasets to select a smaller range to reconstruct.
Useful when the data is huge and you just want to optimise some parameters in a small time frame.

Includes updated tests.


aaron-parsons:
This includes a comprehensive test which passes so I will merge this.

PR #223, opened 2019-07-02
First trial to port Ptypy to Python 3

jcesardasilva:
This is my first trial to port Ptypy to Python 3 related to the discussion in Issue #21 
I am using python 3.7 and I have translated the code syntax to python3 and changed some packages such as `StringIO.StringIO` which became `io.StringIO`, among other changes. 
This is not ready for merge yet (don't merge), but it may be a beginning for discussing the port and improving it for the final merge. There are some problems that seem to come from the utils.descriptor.

### DIFF
--- a/ptypy/core/ptycho.py
+++ b/ptypy/core/ptycho.py
@@ -128,7 +128,7 @@ class Ptycho(Base):
       are relative paths will be relative to this directory.
 
     [io.rfile]
-    default = "recons/%(run)s/%(run)s_%(engine)s_%(iterations)04d.ptyr"
+    default = "recons/%%(run)s/%%(run)s_%%(engine)s_%%(iterations)04d.ptyr"
     type = str
     help = Reconstruction file name (or format string)
     doc = Reconstruction file name or format string (constructed against runtime dictionary)
@@ -176,7 +176,7 @@ class Ptycho(Base):
     lowlim = -1
 
     [io.autosave.rfile]
-    default = "dumps/%(run)s/%(run)s_%(engine)s_%(iterations)04d.ptyr"
+    default = "dumps/%%(run)s/%%(run)s_%%(engine)s_%%(iterations)04d.ptyr"
     type = str
     help = Auto-save file name (or format string)
     doc = Auto-save file name or format string (constructed against runtime dictionary)
@@ -194,7 +194,7 @@ class Ptycho(Base):
     doc = If ``True`` the current reconstruction will be plotted at regular intervals. 
 
     [io.autoplot.imfile]
-    default = "plots/%(run)s/%(run)s_%(engine)s_%(iterations)04d.png"
+    default = "plots/%%(run)s/%%(run)s_%%(engine)s_%%(iterations)04d.png"
     type = str
     help = Plot images file name (or format string)
     doc = Plot images file name (or format string).
@@ -672,7 +672,7 @@ class Ptycho(Base):
             self.init_engine()
             self.runtime.allstart = time.asctime()
             self.runtime.allstop = None
-            for engine in list(self.engines.values()):
+            for engine in self.engines.values():
                 self.run(engine=engine)
 
     def finalize(self):
@@ -748,11 +748,11 @@ class Ptycho(Base):
             P = Ptycho(content.pars, level=1)
 
             logger.info('Attaching probe and object storages')
-            for ID, s in list(content['probe'].items()):
+            for ID, s in content['probe'].items():
                 s['owner'] = P.probe
                 S = Storage._from_dict(s)
                 P.probe._new_ptypy_object(S)
-            for ID, s in list(content['obj'].items()):
+            for ID, s in content['obj'].items():
                 s['owner'] = P.obj
                 S = Storage._from_dict(s)
                 P.obj._new_ptypy_object(S)
@@ -772,7 +772,7 @@ class Ptycho(Base):
 
             logger.info('Regenerating exit waves')
             P.exit.reformat()
-            P.model._initialize_exit(list(P.pods.values()))
+            P.model._initialize_exit(P.pods.values())
 
         if load_data:
             logger.info('Loading data')
@@ -845,7 +845,7 @@ class Ptycho(Base):
                 # self.exit.clear()
 
                 try:
-                    for pod in list(self.pods.values()):
+                    for pod in self.pods.values():
                         del pod.exit
                 except AttributeError:
                     self.exit.clear()
@@ -863,14 +863,14 @@ class Ptycho(Base):
                             'and runtime')
                 dump = u.Param()
                 dump.probe = {ID: S._to_dict()
-                              for ID, S in list(self.probe.storages.items())}
-                for ID, S in list(self.probe.storages.items()):
+                              for ID, S in self.probe.storages.items()}
+                for ID, S in self.probe.storages.items():
                     dump.probe[ID]['grids'] = S.grids()
 
                 dump.obj = {ID: S._to_dict()
-                            for ID, S in list(self.obj.storages.items())}
+                            for ID, S in self.obj.storages.items()}
 
-                for ID, S in list(self.obj.storages.items()):
+                for ID, S in self.obj.storages.items():
                     dump.obj[ID]['grids'] = S.grids()
 
                 try:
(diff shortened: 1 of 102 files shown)


bjoernenders:
Hi Julio,
thanks for the pull request. This looks like an automatic conversion, is it not? I tried this once myself and it made excessive use of `list()` casts there too, even though we are mostly using `.iteritems()` and `.items()` in a for loop. Basically, with few exceptions, we only need to replace  `.iteritems()` with `items`. Also the `print()` conversion seems fishy to me as it sometimes adds parentheses even though we used the `print()` function. Since we would have to undo many of the changes commited here manually, we might want to consider a different start for the real thing. Maybe like this:

- Use one of the automated scripts to transform the code (e.g. futurize) but capture changes in a patch file.
- Remove all changes that involve `list()` and `print()` except for where it makes sense (which are few cases)
- Apply patch and commit. Then we use that branch as a starting point.

@aaron-parsons @pierrethibault what do you think?

jcesardasilva:
Hi @bjoernenders 
Thank you for looking at it. I started indeed from an automatic translation, but then I faced a couple of  problems. Specially with the configparser (rawConfigParser is deprecated) and the stringIO. I also change the access to dataset in H5py parts, since we should use `dataset[()] `to extract value rather than `dataset.value` as before. 

I also checked that, in Python 3, `items()` now return iterators, and a list is never built. The `iteritems()` method has also gone away, since `items()` in Python 3 works like `viewitems()` in Python 2.7. Maybe my Python knowledge is not so high as your, but this is what I found in Stackoverflow. I use the list cast in my code, but I am willing to learn how to implement this in a better way. 

I tried to switch as much as I could the print to print(), but I did not implement the format style, e.g., `print('This is a str for {} v. {}'.format('ptypy',0.4))`, because it would take a lot of time sorting which is %d, %s, %e, %g or any decimal precision. Indeed the printing out can be optimized.

Anyhow, as I said before, my goal here is to make a start point for further development in the port to Python3. We really need to move to Python3 as some dependencies will soon (if not already) not be upgradable anymore in Python2. 
Any help is very welcome.

jcesardasilva:
Please, don't merge yet. Work on progress....

I removed the entries list() that I added before and which were not necessary. I had also to change the parameters (in docstring and the files in resources) with '%' to avoid interpolationsyntaxerror of the configparser. If you have a better a way to do it, please let me know. By now, we are now capable to import ptypy and use most of the functions. I also remove the parts that were discriminating string from unicode, as everything is unicode in Python3. However, it still presents some problems:

- Cannot find the packages in Experiment: it says the reason is "Reason: No module named XXX", where the XXX is the module name. I have the feeling it has to do with relative imports, but could not find a way to solve. Although it complains when import ptypy, we can still use function of ptypy. 

- Cannot run scripts in templates, for example, this `python ptypy_i13_AuStar_nearfield_9p7keV.py` does not work. Get the the errors below which does not appear in python2.

Any help is welcome here. 

```
Could not import experiment Hdf5Loader from hdf5_loader, Reason: No module named 'hdf5_loader'
Could not import experiment cSAXSScan from cSAXS, Reason: No module named 'cSAXS'
Could not import experiment Savu from savu, Reason: No module named 'savu'
Could not import experiment makeScanPlugin from plugin, Reason: No module named 'plugin'
Could not import experiment ID16AScan from ID16Anfp, Reason: No module named 'ID16Anfp'
Could not import experiment AMOScan from AMO_LCLS, Reason: No module named 'AMO_LCLS'
Could not import experiment DiProIFERMIScan from DiProI_FERMI, Reason: No module named 'DiProI_FERMI'
Could not import experiment FliSpecScanMultexp from optiklabor, Reason: No module named 'optiklabor'
Could not import experiment UCLLaserScan from UCL, Reason: No module named 'UCL'
Could not import experiment NanomaxStepscanMay2017 from nanomax, Reason: No module named 'nanomax'
Could not import experiment NanomaxStepscanNov2016 from nanomax, Reason: No module named 'nanomax'
Could not import experiment NanomaxFlyscanJune2017 from nanomax, Reason: No module named 'nanomax'
Could not import experiment ALS5321Scan from ALS_5321, Reason: No module named 'ALS_5321'
Could not import experiment Bragg3dSimScan from Bragg3dSim, Reason: No module named 'Bragg3dSim'
/mntdirect/_data_id16a_inhouse1/sware/pyvenvid16a/py3id16a/lib/python3.6/site-packages/ptypy/utils/scripts.py:468: DeprecationWarning: object of type <class 'float'> cannot be safely interpreted as an integer.
  True)
/mntdirect/_data_id16a_inhouse1/sware/pyvenvid16a/py3id16a/lib/python3.6/site-packages/ptypy/core/classes.py:126: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  % (type(self), self.ID, owner))
Failed registering instance of <class 'ptypy.core.ptycho.Ptycho'> with ID Ptycho to object None
scans                                              scan00               INVALID
engines                                            engine00             INVALID
Traceback (most recent call last):
  File "ptypy_i13_AuStar_nearfield_9p7keV.py", line 107, in <module>
    P = Ptycho(p,level=5)
  File "/mntdirect/_data_id16a_inhouse1/sware/pyvenvid16a/py3id16a/lib/python3.6/site-packages/ptypy/core/ptycho.py", line 311, in __init__
    defaults_tree['ptycho'].validate(self.p)
  File "/mntdirect/_data_id16a_inhouse1/sware/pyvenvid16a/py3id16a/lib/python3.6/site-packages/ptypy/utils/descriptor.py", line 1001, in validate
    raise RuntimeError('Parameter validation failed:\n  ' + '\n  '.join(raise_reasons))
RuntimeError: Parameter validation failed:
  scans - scan00
  engines - engine00

```

ptypy/core/ptycho.py
```diff
@@ -176,7 +176,7 @@ class Ptycho(Base):
     lowlim = -1
 
     [io.autosave.rfile]
-    default = "dumps/%(run)s/%(run)s_%(engine)s_%(iterations)04d.ptyr"
+    default = "dumps/%%(run)s/%%(run)s_%%(engine)s_%%(iterations)04d.ptyr"
```
jcesardasilva, line 179: Had to replace % by %% to avoid InterpolationSyntaxError in ConfigParser. I don't know if there is better ways.

ptypy/resources/parameter_descriptions.csv
```diff
@@ -11,11 +11,11 @@ level,name,default,type,static,lowlim,uplim,choices,shortdoc,longdoc,userlevel,v
 **NOT IMPLEMENTED**",2,
 0,io,,Param,yes,,,,Global parameters for I/O,,,
 1,home,./,dir,yes,,,,Base directory for all I/O,home is the root directory for all input/output operations. All other path parameters that are relative paths will be relative to this directory.,1,
-1,rfile,recons/%(run)s/%(run)s_%(engine)s_%(iterations)04d.ptyr,str,yes,,,,Reconstruction file name (or format string),Reconstruction file name or format string (constructed against runtime dictionary),1,
+1,rfile,recons/%%(run)s/%%(run)s_%%(engine)s_%%(iterations)04d.ptyr,str,yes,,,,Reconstruction file name (or format string),Reconstruction file name or format string (constructed against runtime dictionary),1,
```
jcesardasilva, line 14: Changed % to %% again because of ConfigParser

ptypy/resources/parameter_descriptions.configparser
```diff
@@ -333,7 +333,7 @@ type = int
 userlevel = 1
 
 [scan.data.chunk_format]
-default = .chunk%02d
+default = .chunk%%02d
```
jcesardasilva, line 336: because configparser

ptypy/resources/parameter_descriptions.configparser
```diff
@@ -51,7 +51,7 @@ type = dir
 userlevel = 1
 
 [io.rfile]
-default = recons/%(run)s/%(run)s_%(engine)s_%(iterations)04d.ptyr
+default = recons/%%(run)s/%%(run)s_%%(engine)s_%%(iterations)04d.ptyr
```
jcesardasilva, line 54: because configparser

alexbjorling:
Some very minor contributions here, now it imports without complaints, but the templates don't run still.

alexbjorling:
`minimal_prep_and_run` runs now, but the live plotter crashes, and the `ptyr` file at the end doesn't look much like a bunch of flowers...

alexbjorling:
Sorry Julio, I ended up starting over from the beginning in #231, but only because I understood a lot more when working through your draft!

aaron-parsons:
I'm going to close this now since we already did the python 3 based off of this branch.

PR #224, opened 2019-07-02
Upgraded function for NFP at ID16A beamline at ESRF and removal of deprecated function

jcesardasilva:
I've upgraded the function ID16Anfp.py for NFP experiments at ID16A beamline at ESRF to take into account the new HDF5 format of the acquired data. 
I also remove the deprecation warning in math_utils.py and in h5rw.py and extended the steps in plot_utils.py

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,12 +1,12 @@
 PTYPY - Ptychography Reconstruction for Python
 ==============================================
 
-|doi|
+|ptypysite|
 
 Welcome Ptychonaut!
 -------------------
      
-|ptypy| [#pronounciation]_ is a
+|ptypy| [#pronounciation]_ [#ptypypaper]_ is a
 framework for scientific ptychography compiled by 
 P. Thibault and B. Enders and licensed under the GPLv2 license.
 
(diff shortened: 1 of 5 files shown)


bjoernenders (commented):
Hi Julio,
thanks for this pull request. I'm happy to merge, but the docstrings need to comply with the numpy format. See here: https://numpydoc.readthedocs.io/en/latest/format.html. For doc generation, we are using the napoleon extension, so docstrings should be compatible with that, too. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html  I don't know if these two are synced well, so don't go crazy.

jcesardasilva:
Hi @bjoernenders 
Thank you for calling my attention to it and I apologize for not having done so before. I tried to look at the other scripts in experiments, but did not find much docstring. So I quickly looked at PtyScan class and tried to make something close. I hope it is ok like this, otherwise, please let me know. See commit 9cd1988: https://github.com/ptycho/ptypy/pull/224/commits/9cd19887df520a259c391998a65f045581227411

PR #226, opened 2019-07-17
BUGFIX:energy, distance and pixel size loaded from file were not added to the metadata.

aaron-parsons:


aaron-parsons:
Again. Small changes so I will merge.

PR #231, opened 2019-08-20
Second attempt at Python3

alexbjorling:
I ended up re-doing @jcesardasilva's work on Python 3, because I learned some things working on his PR #223 today. As the commits show, the procedure was:

1. Running `futurize` stage1 and making sure 2.7 tests still ran
2. Running `futurize` stage2 and noting that everything fails
3. Removing wrappers for 2.7 compatibility (which means choosing `/` or `//` everywhere)
4. Removing unnecessary casts
5. Fixing things as they show up, like explicit checks for `iteritems`, API changes, etc.

At this point `minimal_prep_and_run.py` works and gives the correct numerical result. Next up is

6. To go through the 55 out of 256 tests that fail, making them run
7. Fixing the live plotter if that didn't happen by itself during (6)
8. Generally trying to use everything


alexbjorling:
By the way I assumed that nobody wanted to have 2/3 compatibility (which means using a lot of ugliness, `__future__`, `past`, `builtins`, etc)

alexbjorling:
Ok guys, now all tests run in python 3 except three, which are all about writing pickled data with `h5py`. Turns out that `h5` files can't store the null character in strings, and for some reason we get lots of these with pickle whereas `cPickle` didn't use to produce them.

The pickle/h5 problem sounds like it's in @aaron-parsons's area of expertise. Certainly outside mine...

Otherwise, any comments on this?

pierrethibault:
Hi Alex, great job, thanks! I believe that the pickle encoder in h5rw can be safely removed. It was meant as a fallback for unknown objects, but I don't think it gets ever used in ptypy. Plus, pickling is not a safe or portable solution, so it's a good idea to remove it.

bjoernenders:
Hi @alexbjorling and @jcesardasilva . Great job from both of you guys. I worked my way through the commits and they are good. Well done. I agree that we might want to remove pickling of objects and instead skip them while putting out a warning, that an object was not put into the hdf file, but let's wait for Aarons opinion. He is on vacation currently, so that might take a while.

alexbjorling:
Ok the last commit is more a suggestion, and all tests pass with Python 3.7. There are `np.recarray` values in `Container` objects so they were pickled all along and are now missing from the resulting `ptyr` files.

bjoernenders:
Hum it would be good if the `recarray` be captured in the `.hdf`. Contains lots of important information.

alexbjorling:
- @bjoernenders right. Then we had better dig into them like with ‘Param‘ , and convert to ’dict’ or so. But I have to admit I've never seen a ’recarray’ in real life, I don't really know how they work.

alexbjorling:
@aaron-parsons how can we start testing python3 code on travis without messing up the master branch tests? can that be set up i parallel somehow?

alexbjorling:
And all the rest (@jcesardasilva @pierrethibault @aaron-parsons @bjoernenders et al) how does one go about releasing? Intregrate all new features, release 0.4 as the last python2 version, then merge the (updated) python3 branch and release directly as 0.5?

cmkewish:
Why not make a totally separate git repo: ptycho/ptypy3 (?) 
Unless that would make feature/upgrade merging totally unworkable...

jcesardasilva:
Dear colleagues, thank you very much for the work done on the transition to Python3. Special thanks to @alexbjorling for correcting my mistakes and to have gone much further than me on this. IHMO, I really believe that Ptypy should move definitively to Python3 (Python 3.6 +) and the version for python2 to be only a branch or similar for backup solution only. For this to happen, we need to be sure that the Python3 version is fully operational, to decide whether we capture the `recarray` into the .hdf and to comply with travis-ci. Who has experience with the `recarray` to help us with this? Is there any pre-defined schedule for the next Ptypy release?

aaron-parsons:
Hi folks, thanks for all the work on this. I'm just back from holiday but will look at travis and the recarray tomorrow. We also need to check the doc builds. Personally, once we are happy, I'd propose moving straight across to python3 with minimal delay.

The owners of the PR's are then responsible for converting them to python3 and rebasing, but this should be trivial, and we can help if needed.

bjoernenders:
>     * @bjoernenders right. Then we had better dig into them like with ‘Param‘ , and convert to ’dict’ or so. But I have to admit I've never seen a ’recarray’ in real life, I don't really know how they work.

Hi @alexbjorling . The `recarray` is just a "view" on a standard Numpy array which was named with a custom, named datatype. Hence, the `recarray` should be compatible with h5py and I did some testing initial testing that worked just fine. I don't actually know why the recarray was pickled in its io treatment. Anyway, what I'm saying is that it should be possible to save the recarrays as normal arrays and then just cast them as recarray on reading. 

Other option would be that in core.classes we just don't cast them to recarray type.

aaron-parsons:
OK. I've done the build. Now working on reproducing the recarray issue with unittests. I agree with Bjoern. Since h5rw only needs to undo itself, it doesn't really matter too much how it is saved as long as we can undo the process neatly.

alexbjorling:
Thanks Aaron, suddenly green!

aaron-parsons:
@alexbjorling do you have an example where the storing of the recarrray failed? Currently, just using the .dumps() method on the array works for me, but I'm not doing any comprehensive testing, just using the reinstated unit tests. This just uses pickle, which I then convert to an array and store as a special type. What were you using?

aaron-parsons:
Ah don't worry. I see the problem now.

aaron-parsons:
Ok. I fixed the recarray saving. It was obvious when I understood what was happening.

For reference: When you save a string to an hdf5 file using handle[name] = "my happy string", it automatically stores this as a variable length string. Since it doesn't know the length of the string (it's variable), it delimits using a null byte. This is all fine and happy until you save a byte string with null bytes in it. Then it has no idea how to interpret the length of the string. All that was needed was to dump the recarray to a byte string, and then cast to a numpy fixed length string (np.string_).

Next up is to see what state the docs are in. I might try and make a travis build for it to at least report its state on PR's etc...

aaron-parsons:
I've also reinstated the tests + operations of pickling unknown types. It may be we want to get rid of this in the future, but in the interest of keeping things the same and just switching to python3 I think it's best that we don't change this now.

aaron-parsons:
Ok. Well the documentation build categorically does not pass... I'll keep looking into it this afternoon.
build succeeded, 1835 warnings.

aaron-parsons:
OK. Tracking this through there is something wrong with the formatting in some of the autogenerated .tut files (i.e. ptypyclasses,tut). I see this from working back from the previous published doc and the doc I get locally.

It's interleaving text and tables in a weird way in the tut file. Almost like when you multithread and the prints get out of sync. These files are created by the script2rst script and so the error is either in here or the tutorials.

aaron-parsons:
@bjoernenders Just because you might wonder why I did the latest commit, its because in python 3 __idiv__ doesn't exisit. This meant that the ownengine tutorial failed. Instead now you have to use __truediv__ for our application.

aaron-parsons:
Doc now passes with 3 warnings. @bjoernenders I think these are unavoidable. 
**They are:**
"build/lib/ptypy/utils/parallel.py:docstring of ptypy.utils.parallel.MPIrand_normal:56: WARNING: Footnote [1] is not referenced."

This is because we have imported the doc string from np.random.normal and concatenated it to the existing docstring, which I think is the right thing to have done.

"/dls/science/users/clb02321/DAWN_nightly_broke/ptypy_python3/doc/index.rst: WARNING: document isn't included in any toctree"

and

"WARNING: html_static_path entry '/dls/science/users/clb02321/DAWN_nightly_broke/ptypy_python3/doc/_static' does not exist"

Which are both just due to the design of the docs.

I'm just working through the templates now, I noticed some were failing.

aaron-parsons:
OK. About 50 % are failing.Will fix and then I guess we are good to go.

aaron-parsons:
Right. Docs all pass, and most of the templates too. Tests all pass locally, but I'll let travis be the judge.

The two templates that don't pass are:
1. pars_few_alldoc.py  which I have no idea what it is supposed to do.

2. position_evaluation.py which I can't recall working after that PR was merged.

I think that running all the tests and these templates is as much coverage as we could reasonably ask to get without releasing to our lovely community for a play :-).

Feel free to release.and push the doc to gh_pages once the CI comes back green...

aaron-parsons:
@alexbjorling @daurer @jcesardasilva @pierrethibault 

@bjoernenders  and @aaron-parsons just talked about this.

This is what we are left to do:

- [x] 1. Test compatibility 2-->3, 3-->2 for the h5rw library.
- [x] 2. Check the mpi version.
- [x] 3. Draft release notes.
- [x] 4. Aim for release next week.

aaron-parsons:
Going into the templates folder and running single node mpi jobs of all the templates works with openmpi/3.0.0

aaron-parsons:
We can now load old ptyr files into this new version of ptypy.

aaron-parsons:
OK I think this is good to go now. @pierrethibault @bjoernenders I have modified the release notes slightly. Let me know what you think.

aaron-parsons (approved):
This passes all tests that we can realistically do pre-release.

aaron-parsons:
The next things to be done are:
1. merge into master
2. tag and release
3. upload the docs to gh_pages.

alexbjorling:
How is this going guys? Can I help in any way?

@aaron-parsons @bjoernenders @pierrethibault

aaron-parsons:
Hi @alexbjorling ,

It's all ready. We'll be releasing on the 30th October. Bjoern and I have been on holiday.

Aaron

bjoernenders:
Hi @alexbjorling @pierrethibault @aaron-parsons , 
I squash merged this branch into master, meaning all commits here have become a single commit on the master branch. I choose this route becasue this branch was entirely dedicated to the Python 3 transition with no single commit worthy of being revisited and it's safe because nobody branched of the python 3 branch. We can keep this branch for documentation purposes, but github advises to delete this branch now. Let me know your thoughts. It can still be undone. The doc is already up-to-date. Thank you all for the hard work here in particular @alexbjorling @aaron-parsons @jcesardasilva

alexbjorling:
Yes, thanks guys for pushing this through, it's easy to make a working port but the details (testing docs, etc) are hard work. Looking forward to clearing up the mess of python2/python3 environments and scripts that set PYTHONPATH manually all over our beamline.

I think the squash was the right choice and that the branch should be deleted.

alexbjorling:
@bjoernenders shall I just delete it? edit: I did

I suppose we should also close #223 and delete that branch.

PR #232, opened 2019-08-30
Synchronize all positions before saving (relevant for position refinement)

daurer:
I needed this small tweak to make sure all updated positions are stored in the final .ptyr file. Feel free to suggest better implementation before considering to merge into master.

### DIFF
--- a/ptypy/core/ptycho.py
+++ b/ptypy/core/ptycho.py
@@ -906,7 +906,7 @@ class Ptycho(Base):
                 minimal.positions = {}
                 for ID, S in self.obj.storages.items():
                     minimal.obj[ID]['grids'] = S.grids()
-                    minimal.positions[ID] = np.array([v.coord for v in S.views])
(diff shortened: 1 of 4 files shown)


ptypy/core/ptycho.py
```diff
@@ -906,7 +906,7 @@ def save_run(self, alt_file=None, kind='minimal', force_overwrite=True):
                 minimal.positions = {}
                 for ID, S in self.obj.storages.items():
                     minimal.obj[ID]['grids'] = S.grids()
-                    minimal.positions[ID] = np.array([v.coord for v in S.views])
+                    minimal.positions[ID] = np.array([v.coord for v in S.views if v.pod.pr_view.layer==0])
```
daurer, line 909: @alexbjorling Are you okay with this change? I think it makes sense to only save only one position per actual view on the object, picking the one with layer=0 in the case of using probe modes.

aaron-parsons (approved):

PR #234, opened 2019-09-17
Fixes plot client, adds grids to zmq packet.

aaron-parsons:
@bjoernenders 

I've added axes (from Storage.grids()) to the zmq packet, since I needed it, and I thought it was reasonable for most scientific uses.

Is there any reason I shouldn't do this?


bjoernenders (changes requested):
Grids is too expensive. We should opt for `origin` instead if this merge request is at all needed still

PR #238, opened 2019-11-01
Nanomax classes to match new python3 master

alexbjorling:
Not sure what the release schedule for 0.4 is, but if this could fit in before release, that would be splendid.


ptypy/experiment/nanomax_streaming.py
```diff
@@ -0,0 +1,208 @@
...
+#                    break
+            except zmq.ZMQError:
+                # no more data available - working around bug in ptypy here
+                if self.latest_det_index_received < self.info.min_frames * parallel.size:
+                    print('have %u detector frames, waiting...' % (self.latest_det_index_received + 1))
+                    time.sleep(.5)
+                else:
+                    break
+
+        print('-------------------------------------')
```
bjoernenders: Consider using the logging functionality of ptypy instead of print. Print shoudl be used for debugging.

bjoernenders (approved):
I had a quick look and I only found the repeated use of "print" as something that you might to change to "logger.info" etc. Otherwise, PtyScan classes are the responsibility of the beamlines. So I agree to merge. But it will probably be in the rolling master now.

ptypy/experiment/nanomax_streaming.py
alexbjorling: Busted! You're right, I should use the logger.

daurer:
Was looking through the open PRs and this one looks kind of ready to be merged. @alexbjorling Do you want this merged into master? Only recommendation would be to replace print statements with the logger (as suggested above).

alexbjorling:
Sorry @daurer I missed this update. Hang on, I'll just put in another tiny feature first and then ask to merge.

alexbjorling:
Ok! Please merge and close this request. @daurer @bjoernenders

ptypy/experiment/__init__.py.orig
```diff
@@ -0,0 +1,93 @@
...
+                   ('.AMO_LCLS', 'AMOScan'),
+                   ('.DiProI_FERMI', 'DiProIFERMIScan'),
+                   ('.optiklabor', 'FliSpecScanMultexp'),
+                   ('.UCL', 'UCLLaserScan'),
+                   ('.nanomax', 'NanomaxStepscanMay2017'),
+                   ('.nanomax', 'NanomaxStepscanNov2016'),
+                   ('.nanomax', 'NanomaxFlyscanJune2017'),
+                   ('.ALS_5321', 'ALS5321Scan'),
+                   ('.Bragg3dSim', 'Bragg3dSimScan')]
+>>>>>>> master
```
bjoernenders: Here is a merge conflict.

bjoernenders (changes requested):

ptypy/experiment/__init__.py.orig
bjoernenders: I actually don't understand this file __init__.py.orig. Did you commit a backup file by accident?

alexbjorling: yep sorry, must have backed it up when resolving some earlier conflict

bjoernenders (approved):

bjoernenders:
I would do a squash merge, since this a topic/side branch. Is that ok with you?

alexbjorling:
> I would do a squash merge, since this a topic/side branch. Is that ok with you?

of course!

bjoernenders:
Merged and topic branch deleted.

PR #239, opened 2019-12-23
First implementation of the concept of extension point in engines

pierrethibault:
Do not merge yet. First draft.


bjoernenders:
Currently, we favor inheritance over extension points with customized engines. For the new release there will a folder for it `ptypy/custom`. But we can keep an eye open if this is needed in the future.

PR #243, opened 2020-03-02
Attach float_intens to runtime in ML finalize

daurer:
this is to be able read the ML floating intensities from the .ptyr file.


pierrethibault (approved):
Looks good. Thanks.

PR #244, opened 2020-03-04
Added new metric parameter for position correction

daurer:
This is to enable using the likelihood as a metric for position refinement. It also provides an easy way to implement other metrics in the future.


pierrethibault (approved):
Thanks. Good addition.

PR #245, opened 2020-03-06
Fixed bug in position refinement parameter parsing

### LINKED ISSUE #256: updating defaults in parameter subtree does not work properly
I realised this problem in the case of the position refinement parameters. Becasuse the entire engine parameter tree is updated on the main level, entries in the sub-tree (the position refinement parameters) are not updated indivudally insteead the entire list of parameters is overwritten at once. This means that if the user does not specify all available options, there will be missing attributes because the defaults are no longer available.

daurer:
I found that in the current version of the position correction engine, that defaults and user-provided parameters are not updated properly. Position correction only works if all posref parameters are provided in the configuration. 

I've made a quick fix for this. @pierrethibault is there are more elegant way of solving this?

I've also included new tests.

### DIFF
--- a/ptypy/engines/DM.py
+++ b/ptypy/engines/DM.py
@@ -122,11 +122,6 @@ class DM(PositionCorrectionEngine):
         """
         super(DM, self).__init__(ptycho_parent, pars)
 
-        p = self.DEFAULT.copy()
-        if pars is not None:
-            p.update(pars)
-        self.p = p
-
         # Instance attributes
         self.error = None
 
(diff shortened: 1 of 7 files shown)


bjoernenders:
This is a general issue. It looks like all subclasses read the parameters again. Subclasses shouldn't need to read the parameter tree. Needs to be a separate issue. Thanks for finding this bug Benedikt!

daurer:
Removed the second reading of the parameter tree in DM and ML. But this does not fix the issues with the position refinement subtree, which needs to read sepearately. I have moved that part now into the init() of the PositionCorrection engine.

PR #246, opened 2020-04-10
Power correction and log-likelihood error for ePIE

daurer:
Small improvements for the ePIE engine, not tested with MPI.

### DIFF
--- a/ptypy/engines/ePIE.py
+++ b/ptypy/engines/ePIE.py
@@ -115,6 +115,11 @@ class EPIE(BaseEngine):
     lowlim = 0.0
     help = Pixel radius around optical axes that the probe mass center must reside in
 
+    [compute_log_likelihood]
+    default = True
+    type = bool
+    help = A switch for computing the log-likelihood error
+
     """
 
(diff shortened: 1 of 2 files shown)


bjoernenders (commented):
LL is a switch, just like for DM
Power correction is ok.
Consult with Alex

alexbjorling:
To be honest I have never used this engine, wrote it because some people here were super insistent that they needed ePIE... @daurer I guess you have used it since you are touching it up?

daurer:
@alexbjorling thanks for the info. I have recently started testing ePIE, but more out of curiousity how it compares to the other engines and to other packages. For the data I was testing, it seemd like the probe can easily explode unless its power is constrained in some way. Therefor my suggestion to implement a power correction (similar to what we are doing in DM). 

Anybody else out there using ePIE who would be against this change?

alexbjorling:
What I didn't like was the edge effects between domains which makes the trajectory of the  reconstruction depend on how many processes you use. But great that you're looking into it!

bjoernenders (approved):
Looks good, go ahead

PR #260, opened 2020-06-10
Corrected shebang in scripts and changed Matplotlib backend to tkagg

jcesardasilva:
Corrects the shebang in scripts to be compatible with Python3 and to drop Python2 compatibility. In order to make ptypy.plotclient to run properly, I changed the matplotlib backend from agg to tkagg in case of NODISPLAY in plot_utils.py. When using the backend agg, neither ptypy.plotclient nor ptypy.plot display any plots.

### DIFF
--- a/ptypy/utils/plot_utils.py
+++ b/ptypy/utils/plot_utils.py
@@ -17,7 +17,7 @@ import matplotlib.cm
 # importing pyplot may fail when no display is available.
 NODISPLAY = (os.getenv("DISPLAY") is None)
 if NODISPLAY:
-    matplotlib.use('agg')
+    matplotlib.use('tkagg')
 import matplotlib.pyplot as plt
 
 from .verbose import logger
--- a/ptypy/utils/plot_utils.py
+++ b/ptypy/utils/plot_utils.py
@@ -13,11 +13,6 @@ import os
 from PIL import Image
 import matplotlib as mpl
 import matplotlib.cm
-
-# importing pyplot may fail when no display is available.
-NODISPLAY = (os.getenv("DISPLAY") is None)
-if NODISPLAY:
-    matplotlib.use('tkagg')
 import matplotlib.pyplot as plt
 
 from .verbose import logger
--- a/scripts/ptypy.csv2cp
+++ b/scripts/ptypy.csv2cp
@@ -1,4 +1,4 @@
-#! /usr/bin/python
+#!/usr/bin/env python3
 
 from ptypy import utils as u
 import textwrap
--- a/scripts/ptypy.inspect
+++ b/scripts/ptypy.inspect
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 from ptypy import utils as u
 from ptypy import io
--- a/scripts/ptypy.new
+++ b/scripts/ptypy.new
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 import sys
 import argparse
--- a/scripts/ptypy.plot
+++ b/scripts/ptypy.plot
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 import sys
 import argparse
--- a/scripts/ptypy.plotclient
+++ b/scripts/ptypy.plotclient
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 import sys
 import argparse
--- a/scripts/ptypy.run
+++ b/scripts/ptypy.run
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import argparse
 import json


ptypy/utils/plot_utils.py
```diff
@@ -17,7 +17,7 @@
 # importing pyplot may fail when no display is available.
 NODISPLAY = (os.getenv("DISPLAY") is None)
 if NODISPLAY:
-    matplotlib.use('agg')
+    matplotlib.use('tkagg')
```
bjoernenders: This would establish TK dependency even for non GUI case, which is not intended.

daurer:
@jcesardasilva  thanks for catching the shebang mistakes. What is the problem with the agg backend? For me the plotclient and ptypy.plot work fine as it is now.

jcesardasilva:
Thank you @bjoernenders for the remark. Sorry about not being so clear when explaining my issue. @bjoernenders and @daurer, the problem is that, at least in my computer, `NODISPLAY` is always `none`, so plot_utils.py will put the backend as `agg`. By doing so, I cannot display the images, because I get the following warning: 

```
/usr/local/lib/python3.7/site-packages/ptypy/utils/plot_client.py:658: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  pause(.1)
```

Although this is only a warning, it is enough for not displaying the images. Anyhow, I noticed that in case pause cannot be imported, plot_utils will try to import PyQt4 (which is deprecated), PyQt5, WX or GTK. All this add dependencies too. So, what is your advice? 

I understand that `agg` is for writing files, but I don't know why it is called when I try to display the data only for checking the evolution of the calculation. Is there a way to set NODISPLAY when I call ptypy.plotclient in way that the backend `agg` is not set?

daurer:
@jcesardasilva thanks for the explanation. I wonder why $DISPLAY is not set on your machine even though you probably have graphics available. What operating system are you on? What happens if you do ``export DISPLAY=:0``?

jcesardasilva:
Thank you @daurer for your prompt reply. I do not necessarily have graphics available if I don't request them in the cluster, which is in Debian 8. 
```
jdasilva@hpc2-0701:~$ echo $DISPLAY

jdasilva@hpc2-0701:~$
```
I also have the same problem in MacBook, which uses the Catalina OS. 

```
➜  ~ echo $DISPLAY

➜  ~
```
For sure, it works if I previously set `DISPLAY:=0` as you said. From this, it comes my remark/question/request: could we find a way in which this is managed directly by Ptypy when calling the plotclient? Thank you very much in advance.

daurer:
I don't think that Ptypy should overwrite the $DISPLAY variable when calling the plotclient, because it is not obvious what it should be (could be ":0" or ":1" or something else). @bjoernenders  what do you think?

@jcesardasilva do you have Xquartz installed on your Mac? And how you do run the plotclient on the cluster? Over ssh using X forwarding or some other way? Maybe you can include the ``export DISPLAY=:0" in your bashrc or write a small wrapper that sets the DISPLAY and then calls the plotclient?

jcesardasilva:
Thank you @daurer for the suggestions. I will try to answer your question. Please, don't hesitate to comment if it is not clear.

1) Since I moved to OS Catalina, which is entirely 64 bits, I dropped XQuartz and X11. So, I don't have XQuartz installed in MacBook computer. 

2) In the cluster, I use `ssh -X` indeed to connect to the frontend machines, but to get the HPC machines, where I can do calculations, I use the batch scheduler OAR (https://oar.imag.fr), which soon we will move to the workload manager SLURM (https://slurm.schedmd.com/documentation.html). In any case, I can know exactly in which machines the calculations are running and I can access that machine to call `plotclient` to check the evolution of the calculations. 

3) I apologize, but I think I need to disagree with the idea of modifying the bashrc or write a wrapper to set the DISPLAY. This is not general, and would need to be included in the Ptypy documentation for people who may face the same problem as me. 

Could we discuss other solutions? 

Do we still need the check of the `NODISPLAY` at the beginning of the `plot_utils.py`? The comment says: `# importing pyplot may fail when no display is available.` Is it still the case today with the new versions of the Matplotlib and python 3+? I have DISPLAY not set and I can import `matplotlib.pyplot as plt` without problem.

In case we want to keep the check of the `NODISPLAY`, could we put the check within a `try/except` block? In principle, if I remove the check, the code runs smoothly in my machines.

daurer:
@jcesardasilva  Thanks for your answers. I still don't fully understand how your system works without DISPLAY being set, but I am also no longer a mac user and maybe things have changed :) 

I agree with you, that we might not need the NODISPLAY part anymore in the plot_utils.py, as you say it seems that more recent versions of matplotlib (and probably python 3) are able to automatically select the appropriate backend. So, if @bjoernenders agree's we can probably just delete those lines.

ptypy/utils/plot_utils.py
daurer: I would suggest deleting those lines, as with recent versions of matplotlib import pyplot seems to work also in the non-GUI case.

jcesardasilva: I was no sure I should already propose the pull request to implement this, but I decided to do so. Please, let me know if this is ok with you. Thanks.

daurer (approved):
I am okay with these changes, as they seem to work fine/better in modern python 3 environments (incl. matplotlib)

bjoernenders:
Ok that's fine then with me. But I'd rather have the lines uncommented instead of deleted. For future reference and in case we need to reinstate it.

PR #266, opened 2020-08-19
PYCUDA smoothing kernel

daurer:
* reimplemented gaussian smoothing kernel for PYCUDA
* wrote new UNITY tests, all passing (although had to lower tolerance to 1e-4 for some of them)
* made DM_pycuda and DM_pycuda_streams use the new kernel for smoothing the object -> tested and seems to work
* fixed ML_pycuda to use new kernel for smoothing the gradient -> tested and seems to work
* few more fixes to avoid some warnings
* moved context creation from init() to engine_initialize(), this way we can use a new context for each pycuda engine.

### DIFF
--- a/ptypy/engines/DM_pycuda.py
+++ b/ptypy/engines/DM_pycuda.py
@@ -19,7 +19,7 @@ from ..utils import parallel
 from . import BaseEngine, register, DM_serial, DM
 from ..accelerate import py_cuda as gpu
 from ..accelerate.py_cuda.kernels import FourierUpdateKernel, AuxiliaryWaveKernel, PoUpdateKernel, PositionCorrectionKernel
-from ..accelerate.py_cuda.array_utils import ArrayUtilsKernel
+from ..accelerate.py_cuda.array_utils import ArrayUtilsKernel, GaussianSmoothingKernel
 from ..accelerate.array_based import address_manglers
 
 MPI = parallel.size > 1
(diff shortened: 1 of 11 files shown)


ptypy/engines/DM_pycuda.py
```diff
@@ -361,20 +361,16 @@ def object_update(self, MPI=False):
...
-                queue.synchronize()
-                obj_gpu *= cfact
-                print 'gauss: '  + str(time.time()-t2)
-            else:
-                obj_gpu *= cfact
-            """
-            cfact = self.ob_cfact[oID]
+                smooth_mfs = [self.p.obj_smooth_std, self.p.obj_smooth_std]
+                ob_gpu_tmp = gpuarray.empty(ob.shape, dtype=np.complex64)
+                self.GSK.convolution(ob.gpu, ob_gpu_tmp, smooth_mfs)
```
bjoernenders, line 370: Does the kernel also work in-place?

bjoernenders (approved):
This looks very good so far. I wonder why the rtol needs to be as high as 1e-04? Shouldn't stop us from merging but I'm curious

ptypy/engines/DM_pycuda.py
daurer: No, it currently does not work in-place and I don't really see how it could be done in an efficient way. We could have some kind of local cache I guess, but then we probably need some extra sync between threads to know when its safe to overwrite the input. But maybe you have a clever suggestion?

daurer:
Thanks Bjoern! I am going to merge this back into the pycuda branch now. Yes, we should definitley check why rtol=1e-4 is required, it could be something related to the defintion/implementation of the gaussian function.

PR #267, opened 2020-09-02
WIP: More error metrics for pycuda engines

daurer:
* Implemented log_likelihood metric for DM_pycuda and DM_serial with new kernel
* Implemented exit_error metric for DM_pycuda and DM_serial with new kernel


ptypy/accelerate/py_cuda/kernels.py
```diff
@@ -129,6 +130,36 @@ def fourier_update(self, f, addr, fmag, fmask, mask_sum, err_fmag, pbound=0):
...
+                                 mask,
+                                 mag,
+                                 addr,
+                                 ferr,
+                                 np.int32(self.fshape[1]),
+                                 np.int32(self.fshape[2]),
+                                 block=(32, 32, 1),
+                                 grid=(int(mag.shape[0]), 1, 1),
+                                 stream=self.queue)
+        self.error_reduce(addr, err_phot)
```
bjoernenders, line 147: I wonder if it's not better to have the reduction step outside the log_likelihood() call. It would be more in line with the other reductions, would it not?

daurer: There are other examples where the reduction call is inside the main call of the kernel, e.g. fill_b and fill_b_reduce and also in the floating_intensity kernel. But I am okay with both. But maybe we should decide one way and then be consistent about it. Which one would you prefer? Having all reduce kernel calls as seperate python class method or attached to the main class method?

bjoernenders: Ok I was fearing we were being inconsistent. No it's fine keep it as is. It's a minor thing. Maybe flag it as a todo, so we keep track. Feel free to merge.

PR #273, opened 2020-11-10
Allow for negative propagation in geometry.py

bjoernenders:
Removed lower limit of 0 from `distance` parameter to allow for negative propagation.

### DIFF
--- a/ptypy/io/h5rw.py
+++ b/ptypy/io/h5rw.py
@@ -443,7 +443,7 @@ def h5read(filename, *args, **kwargs):
 
     def _load_str(dset):
         #return str(dset[()])
-        return dset.asstr()
+        return dset.asstr()[0]
 
     def _load_unicode(dset):
         return dset[()].decode('utf-8')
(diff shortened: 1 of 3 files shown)


bjoernenders:
@daurer I put in an ugly fix to comply with h5py>=3.0.0. Since variable length strings now find their way into h5py, I wonder if we need to deprecate the _store_string function. Let me know what you think

daurer:
@bjoernenders thanks for the fix. I think we should keep it as it is for now (many users will still be on h5py <3.0.0), but maybe flag it for a future redesign? I am okay if we move to h5py >=3.0.0 requirement for the next major release.

ptypy/io/h5rw.py
```diff
@@ -442,8 +442,11 @@ def _load_ordered_dict(dset, depth):
         return d
 
     def _load_str(dset):
-        return str(dset[()])
-
+        if h5py.version.version_tuple[0]>2:
+            return dset[()].decode('utf-8')
```
daurer, line 446: here, you could reuse ```_load_unicode(dset)```

daurer (approved):

Issue #165, opened 2018-08-03
DMGpu engine doesn't work with 'nearfield' propagtion


carlosesreis:
Hi guys. First of all thanks for the software! We have been using it with GPUs and it goes really well.
We were trying to use DMGpu engine in a nearfield setup but it is dropping us the following error message:

```
Traceback (most recent call last):
  File "ptypyMinRec.py", line 91, in <module>
    P = Ptycho(pars=param_tree, level=5)
  File "/home/carlos/anaconda2/envs/ptypy_gpu_cuda_merged/lib/python2.7/site-packages/Python_Ptychography_toolbox-0.2.0-py2.7-linux-x86_64.egg/ptypy/core/ptycho.py", line 332, in __init__
    self.run()
  File "/home/carlos/anaconda2/envs/ptypy_gpu_cuda_merged/lib/python2.7/site-packages/Python_Ptychography_toolbox-0.2.0-py2.7-linux-x86_64.egg/ptypy/core/ptycho.py", line 655, in run
    self.run(engine=engine)
  File "/home/carlos/anaconda2/envs/ptypy_gpu_cuda_merged/lib/python2.7/site-packages/Python_Ptychography_toolbox-0.2.0-py2.7-linux-x86_64.egg/ptypy/core/ptycho.py", line 609, in run
    engine.iterate()
  File "/home/carlos/anaconda2/envs/ptypy_gpu_cuda_merged/lib/python2.7/site-packages/Python_Ptychography_toolbox-0.2.0-py2.7-linux-x86_64.egg/ptypy/engines/base.py", line 194, in iterate
    self.error = self.engine_iterate(niter_contiguous)
  File "/home/carlos/anaconda2/envs/ptypy_gpu_cuda_merged/lib/python2.7/site-packages/Python_Ptychography_toolbox-0.2.0-py2.7-linux-x86_64.egg/ptypy/engines/DM_gpu.py", line 155, in engine_iterate
    pre_fft = self.propagator[dID].pre_fft
AttributeError: 'BasicNearfieldPropagator' object has no attribute 'pre_fft'
```

Using the farfield option with yours rspa usafe dataset all works fine. 
Could you please help us with this issue?

Thanks,
Carlos

aaron-parsons:
Hi Carlos,

[...]

PR #275, opened 2020-11-10
[WIP] Enable nearfield propagation for pycuda engines

bjoernenders:
This is the pull request to enable nearfield propagation. Please do not review yet.


bjoernenders:
For issue #165 I'm trying to enable NF for the pycuda branch. Feel free to test it or remove bugs. @daurer @jcesardasilva I can continue Thursday.

jcesardasilva:
Hi @bjoernenders, 

I tried to look at your changes and test the code a bit. I don't know if this helps or not, but in any case feel free to ignore my  commit eb6f630df1c8dd2c56209bb7bf688a815a148617 . I've made some small changes, which helps to make the PyCUDA calculations for Far-field propagation to work. Concerning the near-field propagation, it does not seems to like the lambda functions defined for `self.fw` and `self.bw`. Since a `lambda` function is constructed within a class, I've added the `self` as an argument, but now it complains about the attributes now, as you can see below: 
```
Traceback (most recent call last):
  File "ptypy_NFP_prep_recons_CUDA.py", line 322, in <module>
    P = Ptycho(p, level=5)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/core/ptycho.py", line 365, in __init__
    self.run()
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/core/ptycho.py", line 695, in run
    self.run(engine=engine)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/core/ptycho.py", line 649, in run
    engine.iterate()
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/engines/base.py", line 221, in iterate
    self.error = self.engine_iterate(niter_contiguous)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/engines/DM_pycuda.py", line 218, in engine_iterate
    PROP.fw(aux, aux)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/accelerate/py_cuda/kernels.py", line 66, in <lambda>
    self.fw = lambda self,x: self._fft1.ift(self._fft1.ft(x))
AttributeError: 'GPUArray' object has no attribute '_fft1'
```

daurer:
Hi @jcesardasilva @bjoernenders 

I had a look at the code, the problem is not the ```self``` inside the lambda. In the pycuda version of ptypy, the ```fw``` and ```bw``` methods have two arguments, one for the input array, the other for the output array. This also means that stacking together the ```ft``` and ```ift``` methods to make a nearfield version will most likely need something more than just a single lambda function. 

I could have a go at this tomorrow, but I'll wait for @bjoernenders before making any changes...

EDIT: I have uploaded a new template script that actually uses the Nearfield propagator with the DM pycuda engine.

bjoernenders:
Yeah sorry. I didn't test it at all and knew it was riddled with bugs. I just wanted to get something out so you get the idea, I had to prepare a talk for today.

bjoernenders:
@daurer You are totally right with your observations. If you have capacity to repair it, I happily let you do it. Just wanted to point out that 
`self._fft1.ift(self._fft2.ft(x))` 
was on purpose and not an oversight

jcesardasilva:
Thanks to you both. You have done a great job so far and the near field ptycho (NFP) calculations with GPU + ptypy are so close to working fine that I think it is worth tackling this issue. I can try to help you. 
I have plenty of NFP experimental data waiting for this, for which the CPU calculations are unfortunately still too slow.

daurer:
@bjoernenders  Thanks, this looks better! I've made another change, making sure that ```_fft1``` is only used for forward and ```_fft2``` only for backward FFTs. 

The nearfield template example now runs through without errors, but it gives a lot of np.nan values. Maybe @jcesardasilva you can test it on your data and let us know if you get the expected results. There is still a high chance, that we need to make some more bug-fixes before it gives the same output as regular DM.

bjoernenders:
@daurer. An FFT implementation always delivers the forward and the inverse operation. And since we don't have to worry about optical axis here, we don't have to have spearate FFTs. So `fw` works with just and bw only needs an additional for the forward step, cause it ties the kernel to the post_fft filter
So this can't work

jcesardasilva:
I am having somehow a problem with the template , which, I don't know why, gives me the following error:

```
RuntimeError: Parameter validation failed:
  scans.sim.data - propagation
```

I will try it with real data.

bjoernenders:
@jcesardasilva I would wait until it works for benedikt and he and I have found consensus

daurer:
@bjoernenders sorry, maybe I got this wrong then. Just looked at the "farfield" case where _fft1 is used for forward and _fft2 is used for backward. If what you had before is correct, why do we get this error:

```    PROP.fw(aux, aux)
  File "/home/iat69393/.local/lib/python3.7/site-packages/Python_Ptychography_toolbox-0.4.0-py3.7.egg/ptypy/accelerate/py_cuda/kernels.py", line 61, in _fw
    self._fft1.ift(y,y)
  File "/home/iat69393/.local/lib/python3.7/site-packages/Python_Ptychography_toolbox-0.4.0-py3.7.egg/ptypy/accelerate/py_cuda/cufft.py", line 72, in _ift_ext
    self.fftobj.ifft(input.gpudata, output.gpudata)
RuntimeError: Calling IFFT on a forward-initialised instance
```

jcesardasilva:
Ouch! It seems to me that I am diving into a deeper problem with real experimental data (see error message below). It appears to be memory problem, but the .ptyd file makes 194 MB and I am using the following GPU card: 
image

And here is the error:
```
Save file exists but will be overwritten (force_overwrite is True)
Generating copies of probe, object and parameters and runtime
Saving to /data/id16a/inhouse1/commissioning/comm_17may/ihhc3150/id16a/3B_s1/3B_s1_nfp_tomo_15nm_test_analysis/dumps/3B_s1_nfp_tomo_15nm_subtomo001_0000/3B_s1_nfp_tomo_15nm_subtomo001_0000_None_0000.ptyr
--------------------------------------------------------------------------------
0 Iteration (Overlap) #00:  ----- object update -----
Traceback (most recent call last):
  File "ptypy_NFP_prep_recons_CUDA.py", line 322, in <module>
    P = Ptycho(p, level=5)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/core/ptycho.py", line 365, in __init__
    self.run()
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/core/ptycho.py", line 695, in run
    self.run(engine=engine)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/core/ptycho.py", line 649, in run
    engine.iterate()
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/engines/base.py", line 221, in iterate
    self.error = self.engine_iterate(niter_contiguous)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/engines/DM_pycuda.py", line 256, in engine_iterate
    self.overlap_update(MPI=MPI)
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/engines/DM_serial.py", line 414, in overlap_update
    self.object_update(MPI=(parallel.size > 1 and MPI))
  File "/data/id16a/inhouse1/sware/py3venvslurm/cudavenv/lib/python3.8/site-packages/Python_Ptychography_toolbox-0.4.0-py3.8.egg/ptypy/engines/DM_pycuda.py", line 345, in object_update
    queue.synchronize()
pycuda._driver.LogicError: cuStreamSynchronize failed: an illegal memory access was encountered
PyCUDA WARNING: a clean-up operation failed (dead context maybe?)
cuModuleUnload failed: an illegal memory access was encountered
```

bjoernenders:
I think we need a third FFT because cufft.py is not quite set up as the fft.py or we need to consolidate those first.

daurer:
@bjoernenders Yes, also just realised this. Does it look better now? Still need to test if this works with the template.

ptypy/accelerate/py_cuda/kernels.py
```diff
@@ -1,11 +1,90 @@
...
+            self._fft2 = FFT(aux, self.queue,
+                             post_fft=self._p.ikernel,
+                             inplace=True,
+                             symmetric=True,
+                             forward=True)
+            self._fft3 = FFT(aux, self.queue,
+                             post_fft=self._p.kernel,
+                             symmetric=True,
+                             forward=False)
+            self._fft4 = FFT(aux, self.queue,
```
bjoernenders: don't need _fft4

daurer: Was by mistake :)

daurer:
I am still getting a lot of NaNs with the template script, I will need to check a bit deeper where the problem is.

bjoernenders:
_fft3 cannot have a post_fft filter, that's was wrong

bjoernenders:
Try now @daurer :)

daurer:
Still the same problem. I will have a closer look tomorrow.

daurer:
I think I fixed the NF propagator now. It works with the nearfield templates and the DM_pycuda/ML_pycuda engine. @jcesardasilva can you test it on your data? Make sure that you define p.frames_per_block and make it the same as your total nr. of frames. 

@bjoernenders DM_pycuda_streams is still broken for the nearfield template script. It works with the farfield propagator, no idea what's going on there...

jcesardasilva:
Thank you @daurer for this. I've just tried with real experimental data and it worked very well. I did not compare the results with the results without GPU, but it looks visually the same and much much faster. I will keep working on the data and I let you know if I find any problem. But so far, so good. Great work! Thank you @daurer, thank you @bjoernenders

daurer:
@jcesardasilva That's great! I have found another small bug and now the DM_pycuda_streams engine also works with the NF propagator. If you run into any GPU memory issues (e,g, many data frames) you should be able to switch over to the DM_pycuda_streams engine and you would need to reduce ```frames_per_block``` to something smaller than your total nr. of frames. 

@bjoernenders I think this is ready to be merged

jcesardasilva:
Hi again, 

I am trying to compare the results with and without GPU. Surprisingly, they appear to have a different grey-level range with the same amount of iterations of DM, for example. I understand the GPU calculations are implemented differently relative to the CPU version. 

Could it be that there is a scale factor somewhere in the CUDA version which is not the same as the one in the CPU version?  

I send you below the screenshots of the two calculations I did with 2000 iteration of DM with the same dataset. Besides the phase ramp, you can see the phases are better developed in the CPU results than in the  GPU results. The amplitude grey-level range is nearly the same. The probe does not seem to have the same amount of photons, despite the visual similarities.
I tried with more DM iterations (6000 this time) in the GPU version, but it did not improve, the results are similar as the ones with 2000 iterations DM. 

This is the CPU version:
image

And this is the GPU version:
image

templates/ptypy_i13_AuStar_nearfield_9p7keV.py
```diff
@@ -22,7 +23,7 @@
 sim.energy = 9.7
 sim.distance = 8.46e-2
 sim.psize = 100e-9
-sim.shape = 1024
+sim.shape = 256
```
bjoernenders: This doesn't look right. Why reduce to 256 pixel? Nearfield frames are large

jcesardasilva:
Problem solved! I can get the same results with CPU and GPU. Thank you very much.

PR #279, opened 2020-12-07
Introduce fourier_power_bound parameter

daurer:
Currently, the fourier_relax_factor in DM is being normalised to 

> 0.25 * fourier_relax_factor**2 * max_power_data

this can lead to unexpected behaviour and is not very flexible. For example, in some case it might be better to optmise the power bound to given fixed value (independent of the power in the data). I have therefore introduced a new parameter for this.

I have also improved documentation making things more transparent to the user.


ptypy/engines/DM.py
```diff
@@ -87,11 +87,17 @@ class DM(PositionCorrectionEngine):
     lowlim = 0.0
     help = Weight of the current object in the update
 
+    [fourier_power_bound]
+    default = None
+    type = float
+    help = If rms error of model vs diffraction data is smaller than this value, Fourier constraint is met
+    doc = Set this value higher for noisy data. By default, power bound is calculated using fourier_relax_factor
```
bjoernenders: A little more explanation here as to what to set this value to could be helpful.

bjoernenders (approved):

PR #281, opened 2020-12-09
Fix Gaussian filtering kernel

daurer:
* Use Scipy's Gaussian function, which is properly normalised (also numerically)
* Don't call Gaussian GPU kernel for sigmas < 0.1
* Explicitly return the output, pass by reference does not work for the case where no GPU kernel is called.

### DIFF
--- a/ptypy/accelerate/py_cuda/array_utils.py
+++ b/ptypy/accelerate/py_cuda/array_utils.py
@@ -281,6 +281,4 @@ class GaussianSmoothingKernel:
             
         # TODO: is this threshold acceptable in all cases?
         if (stdx <= 0.1 and stdy <= 0.1):
-            output = input
-
-        return output
+            output[:] = input[:]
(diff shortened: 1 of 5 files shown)


ptypy/accelerate/py_cuda/array_utils.py
```diff
@@ -1,6 +1,6 @@
 from . import load_kernel
 from pycuda import gpuarray
-from ptypy.utils import gaussian
+from scipy.ndimage.filters import _gaussian_kernel1d
```
bjoernenders: Isn't there a public method? I don't know if there is version compatibility for private methods

ptypy/test/accelerate_tests/py_cuda_tests/array_utils_test.py
```diff
@@ -126,58 +126,95 @@ def test_transpose_4D(self):
...
+        mfs = 0.0,0.0
         inp_dev = gpuarray.to_gpu(inp)
         out_dev = gpuarray.empty((11,11), dtype=np.complex64)
 
         # Act
         GS = gau.GaussianSmoothingKernel()
-        GS.convolution(inp_dev, out_dev, mfs)
+        out_dev = GS.convolution(inp_dev, out_dev, mfs)
+        print(inp_dev.ptr)
+        print(out_dev.ptr)
```
bjoernenders: print statements

bjoernenders (approved):

PR #287, opened 2021-02-10
Big merge without acceleration

daurer:
A wide range of changes, some in preparation for upcoming GPU acceleration merge.

- re-introduction of fourier power bound (DM) -> still backwards compatible but we might make fourier_relax_factor deprecated in the future
- full integration of the new block model approach, introducing new paramter frames_per_block
- small changes in core
- moved reformatting out of model and into model manager
- moved tests to root level


ptypy/core/ptycho.py
```diff
@@ -993,3 +1008,134 @@ def plot_overview(self, fignum=100):
                            (slice(0, 1), slice(None), slice(None)),
                            cmap='gray')
             fignum += 1
+
+    
+    def redistribute_data(self, div = 'rect', obj_storage=None):
```
daurer: is this used/finished? maybe it should be removed from this merge?

daurer:
@bjoernenders these are most of the changes that we made over the last year, excluding all GPU acceleration code. I kept all the commits, thinking we can anyway squash all of this into a single merge with master.

ptypy/core/ptycho.py
bjoernenders: This is for the ePIE if I recall correctly. It feels like it should be part of manager.py. Maybe we could just make it a private function instead.

daurer: Right, this seems to be here only by mistake. There is already a similar function inside ePIE.py, looking at the commits it seems like this part slipped through when you made an OpenCL merge. I am going to remove it from core/ptycho.py

bjoernenders: It's one level higher than ePIE on purpose, as I thought it could be used for other engines as well. But if it's the same code we could keep it quarantined to ePIE for now.

daurer: I see, the code is actually quite different. I brought the function back into core/ptycho.py but made it private.

bjoernenders (approved):

daurer:
I identified a corner case that causes problems. When using the BlockModel together with MP and the scenario is such that there is only a single view (or less then nr. of MPI procs) left to process, this last view can end up on a rank != 0 and then during the exit.reformat() things can get stuck because rank=0 is no longer available to do bcast. I have included some comments in the code (see [2fd95aa](https://github.com/ptycho/ptypy/pull/287/commits/2fd95aa235cb6e6d98dbc9b1018a623e67935650)) and the template `minimal_prep_and_run_blockmodel.py` is re-producing the issue. @bjoernenders any idea how to fix this in a simple way?

bjoernenders:
Good find! I guess the easy way might be to wait with the "return in case there are no views" until self.distributed is updated. 

The best solution would be if some of these dynamics were explicit and not implicit, like declaring a container as "distributed" or "shared" to begin with. And the behavior or `reformat` is determined via arguments on the storage level. But this breaks the syntax in places, and we can't do this with this update

PR #288, opened 2021-02-18
Make basic fourier update a true blend between DM and AP

daurer:
With this fix, `alpha=1` is normal DM with full reflection (as it was before) but now `alpha=0` gives Alternate Projections (AP) and any other `0 < alpha < 1` is a mix of DM and AP.


bjoernenders:
While I do approve, this is a "major" change. This is best for a 0.x.0 release. But it's also fairly restricted to this file, so I'd like to keep it open for now.

PR #292, opened 2021-03-01
DM_pycuda_stream fully functional plus extras

bjoernenders:
I tested various scripts but I think this is good now:

- `DM_pycuda_stream` engine is now working again on every script I ran
- `DM_pyucda` and `DM_pycuda_stream` work with a staggered arrival of data "i.e. streaming" and adjust the memory on the GPU dynamically. `DM_pycuda_stream` will saturate GPU until only 200MB are left (like `DM_pycuda_streams`
- Harmonized startup between the engines
- The GPU memory utils are shaped to satisfy the controls flow of a single compute stream with two I/O streams
- I picked Reikna over the CudaFFT now as default, because the cuda fft adds a 24sec delay in startup time (on our Cori cluster) and there was room in the compute stream. Still recons with lots of iterations and lots of data might still benefit from the cuda version.

@daurer could you battle test the engines with your scripts?


daurer (approved):

daurer:
The `DM_pycuda_stream` also works now with my real data scripts. We should probablt have a user parameter in DM_pycuda for switching between "reikna" and "cufft".

PR #294, opened 2021-03-03
Gpu flexible datatypes

xcelerit-team:
All used kernels are transformed to allow flexibility in the data types they use, in order to control/test precision issues.

Note that since there is a common math_type in Python added to the kernel classes, setting this has a cross effect on other GPU kernels. We can refine that once we've determined the best / sufficient data types for all kernels in the tests.

### DIFF
--- a/ptypy/accelerate/cuda_pycuda/cuda/build_aux.cu
+++ b/ptypy/accelerate/cuda_pycuda/cuda/build_aux.cu
@@ -1,24 +1,33 @@
+/** build_aux kernel.
+ *
+ * Data types:
+ * - IN_TYPE: the data type for the inputs (float or double)
+ * - OUT_TYPE: the data type for the outputs (float or double - for aux wave)
+ * - MATH_TYPE: the data type used for computation 
+ */
+
 #include <thrust/complex.h>
 using thrust::complex;
 
 extern "C" __global__ void build_aux(
-    complex<float>* auxiliary_wave,
-    const complex<float>* __restrict__ exit_wave,
+    complex<OUT_TYPE>* auxiliary_wave,
+    const complex<IN_TYPE>* __restrict__ exit_wave,
     int B,
     int C,
-    const complex<float>* __restrict__ probe,
+    const complex<IN_TYPE>* __restrict__ probe,
     int E,
     int F,
-    const complex<float>* __restrict__ obj,
+    const complex<IN_TYPE>* __restrict__ obj,
     int H,
     int I,
     const int* __restrict__ addr,
-    float alpha)
+    IN_TYPE alpha_)
 {
   int bid = blockIdx.x;
   int tx = threadIdx.x;
   int ty = threadIdx.y;
   int addr_stride = 15;
+  const MATH_TYPE alpha = alpha_;   // type conversion
 
   const int* oa = addr + 3 + bid * addr_stride;
   const int* pa = addr + bid * addr_stride;
@@ -35,9 +44,14 @@ extern "C" __global__ void build_aux(
                    // (it will work for less as well)
     for (int c = tx; c < C; c += blockDim.x)
     {
+      // temporaries to convert to MATH_TYPE in case it's different to storage type
+      complex<MATH_TYPE> t_obj = obj[b * I + c];
+      complex<MATH_TYPE> t_probe = probe[b * F + c];
+      complex<MATH_TYPE> t_ex = exit_wave[b * C + c];
+
       auxiliary_wave[b * C + c] =
-          obj[b * I + c] * probe[b * F + c] * (1.0f + alpha) -
-          exit_wave[b * C + c] * alpha;
+         t_obj * t_probe * (MATH_TYPE(1) + alpha) -
+          t_ex * alpha;
     }
   }
 }
--- a/ptypy/accelerate/cuda_pycuda/cuda/build_aux_no_ex.cu
+++ b/ptypy/accelerate/cuda_pycuda/cuda/build_aux_no_ex.cu
@@ -1,23 +1,32 @@
+/** build_aux without exit wave kernel.
+ *
+ * Data types:
+ * - IN_TYPE: the data type for the inputs (float or double)
+ * - OUT_TYPE: the data type for the outputs (float or double - for aux wave)
+ * - MATH_TYPE: the data type used for computation 
+ */
+
 #include <thrust/complex.h>
 using thrust::complex;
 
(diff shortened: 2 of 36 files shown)


ptypy/accelerate/cuda_pycuda/cuda/convolution.cu
```diff
@@ -169,7 +177,7 @@ extern "C" __global__ void convolution_col(const DTYPE *__restrict__ input,
     if (gby + ty >= width || gbx + tx >= height)
         return;
 
-    // compute
+    // compute - will be complex<double> if kernel is double
     auto sum = shm[(tx + kernel_radius) * BDIM_Y + ty] * kernel[0];
```
daurer, line 181: Does this work if DTYPE != MATH_TYPE?

ptypy/accelerate/cuda_pycuda/cuda/build_aux.cu
```diff
@@ -1,24 +1,33 @@
...
+    const complex<IN_TYPE>* __restrict__ probe,
     int E,
     int F,
-    const complex<float>* __restrict__ obj,
+    const complex<IN_TYPE>* __restrict__ obj,
     int H,
     int I,
     const int* __restrict__ addr,
-    float alpha)
+    IN_TYPE alpha_)
```
bjoernenders, line 24: precision of alpha is kinda irrelevant. so assuming that it's a float is fine. double precision input is not needed

ptypy/accelerate/cuda_pycuda/cuda/build_aux.cu
```diff
@@ -1,24 +1,33 @@
+/** build_aux kernel.
+ *
+ * Data types:
+ * - IN_TYPE: the data type for the inputs (float or double)
+ * - OUT_TYPE: the data type for the outputs (float or double - for aux wave)
+ * - MATH_TYPE: the data type used for computation
```
bjoernenders, line 6: I don't understand why this kernel needs 3 types. It should be fine to operate this kernel with a single type

ptypy/accelerate/cuda_pycuda/cuda/build_aux_no_ex.cu
```diff
@@ -1,23 +1,32 @@
+/** build_aux without exit wave kernel.
+ *
+ * Data types:
+ * - IN_TYPE: the data type for the inputs (float or double)
+ * - OUT_TYPE: the data type for the outputs (float or double - for aux wave)
+ * - MATH_TYPE: the data type used for computation
```
bjoernenders, line 6: Same here. This is a simple kernel.

ptypy/accelerate/cuda_pycuda/cuda/error_reduce.cu
```diff
@@ -1,17 +1,24 @@
...
   int tx = threadIdx.x;
   int ty = threadIdx.y;
   int batch = blockIdx.x;
-  extern __shared__ float sum_v[1024];
+  extern __shared__ ACC_TYPE sum_v[1024];
 
   int shidx =
       ty * blockDim.x + tx;  // shidx: index in shared memory for this block
-  float sum = 0.0f;
+  ACC_TYPE sum = ACC_TYPE(0.0);
```
bjoernenders, line 21: I don't see actual typecasting in the accumulation below (`sum+=ferr[idx]`). Is that not need?

ptypy/accelerate/cuda_pycuda/cuda/pr_update2.cu
```diff
@@ -1,3 +1,12 @@
+/** ob_update.
```
bjoernenders: wrong copy paste

ptypy/accelerate/cuda_pycuda/cuda/pr_update2.cu
```diff
@@ -1,3 +1,12 @@
...
+ *
+ * Data types:
+ * - IN_TYPE: the data type for the inputs (float or double)
+ * - OUT_TYPE: the data type for the outputs (float or double)
+ * - MATH_TYPE: the data type used for computation
+ * - DENOM_TYPE: data type for the denominator (double,float,complex<double>,complex<float>)
```
bjoernenders, line 7: MATH type replaces DENOM type below. So why do we need DENOM type? This is starting to get confusing

bjoernenders (commented):
Hi, I guess I'm having trouble for unilaterally inserting a  MATH_TYPE. Like why does it make sense to execute a=b*c+d in a different precision than input and output type? The DENOM_TYPE I also  struggle with. Let's just make the denominator real and of ACC_TYPE/MATH_TYPE.

ptypy/accelerate/cuda_pycuda/cuda/convolution.cu
xcelerit-team: the compiler uses float promotion, i.e. the result is the larger of the 2 floating point types. Auto deduces the return type.

ptypy/accelerate/cuda_pycuda/cuda/pr_update2.cu
xcelerit-team: fixed

ptypy/accelerate/cuda_pycuda/cuda/error_reduce.cu
xcelerit-team: not required, implicit conversion takes care of it.

ptypy/accelerate/cuda_pycuda/cuda/build_aux.cu
xcelerit-team: it needs to match the type passed in from Python. if in type is double, alpha will be double also.

ptypy/accelerate/cuda_pycuda/cuda/build_aux.cu
xcelerit-team: this if for consistency - all kernels will have these types now. And MATH_TYPE is for internal calculations in higher precision.

ptypy/accelerate/cuda_pycuda/cuda/pr_update2.cu
xcelerit-team: this will be fixed when denominator type is always real. At the moment, it can be real or complex, which is why this type is there.

xcelerit-team:
I have now added an ACC_TYPE to the tiled update kernels and a note explaining register spilling effects that may happen if too many modes are used, as discussed.

daurer (approved):

bjoernenders (approved):

PR #298, opened 2021-03-05
Gpu investigate ML floating intensity kernel

daurer:
The `test/accelerate_tests/cuda_pycuda_tests/dls_tests/dls_floating_intensity_accuracy_test.py` test reproduces the problem by comparing the base agains the pycuda kernel using the same input.

Observations so far:
- forward projector causes small nr. of mismatched pixels (0.263 %) for the aux waves. 
- make model as a consequence causes bigger but still small nr. of mismatched pixels (2.02 %) for the aux waves.
- The floating intensity kernel suddenly causes a huge nr. of mismatched pixels (95.9%) for the Imodel.
- Then, without the floating intensity option, the main kernel causes a mismatch 6.92% on the aux waves.
- With the floating intensity option, the main kernel has a mismatch of 88.1% on the aux waves.
- The reduced likelihood error has a mismatch of 100% for the floating intensity option, but 0% (pass) without the floating intensity option.

This behaviour also matches my experience when runninf the full reconstruction. Whithout using the floating intensity option, everything seems to be fine (even though there are inprecisions on the GPU) but when using the floating intensity option, something goes wrong big time.


daurer:
Found a bug in the intens_renorm.cu kernel #299 #301

PR #299, opened 2021-03-08
Potential fix for floating intensity kernel

daurer:
It looks like we should not update fic in-place while it is still used to update Imodel. Have fixed this issue for now by introducing new array for the nominator (fic_nom) and renamed fic_tmp to fic_den, but certainly there must be a way to fix this without the extra memory...

### DIFF
--- a/ptypy/accelerate/base/engines/ML_serial.py
+++ b/ptypy/accelerate/base/engines/ML_serial.py
@@ -395,6 +395,7 @@ class GaussianModel(BaseModelSerial):
                     f["addr"] = addr
                     f["I"] = I
                     f["fic"] = fic
+                    f["Imodel"] = GDK.npy.Imodel
 
             if self.p.floating_intensities:
                 GDK.floating_intensity(addr, w, I, fic)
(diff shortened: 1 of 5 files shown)


bjoernenders:
Good catch. I think that only happens when multiple blocks are needed for the same z-index (i.e. frame index) so it only happens in real life. The old test wouldn't catch it, bc the frames are too small. The additional memory footprint is pretty small. So it's ok from my end. I wonder if we have issues anywhere else, where the test size is so small it fits into a single block on GPU. Maybe we need a test size in between the small tests and performance test that we use for unity testing with base

xcelerit-team:
I had a further discussion with Benedikt this morning and found a better fix for this without an extra array, by having 1 thread block work on one iz, reading fic/fic_tmp into shared memory, and then sync. This way we don't have a race condition with writing back to fic, still have more than enough threads to fill the GPU, and we're also avoiding many threads reading the same values from global memory. I'm following up with a pull request for this shortly and we can probably discard this one then.

daurer:
Closing this PR, because there is a better solution #301

PR #303, opened 2021-03-09
Gpu hackathon: crop pad

xcelerit-team:
This GPU implementation makes all tests pass for crop/pad. However, I feel the tests should be extended to cover the following corner cases / special cases:
- higher dimensionality, like 5D or 6D
- larger arrays, possibly random, where A is not just filled with zeros initially (also to use multiple thread blocks on the GPU and make sure that behaves fine)

We should also add a performance test (skipped by default) with something large, to run a profile and make sure the implementation is efficient.

Note that this kernel is the full fill3D, not just the simplified version - tests for that case should also be added to verify if that's working fine.

### DIFF
--- a/ptypy/accelerate/cuda_pycuda/array_utils.py
+++ b/ptypy/accelerate/cuda_pycuda/array_utils.py
@@ -21,12 +21,6 @@ class ArrayUtilsKernel:
             'ACC_TYPE': 'double' if acc_dtype==np.float64 else 'float',
             'BDIM_X': 1024
         })
-        self.transpose_cuda = load_kernel("transpose", {
-            'DTYPE': 'int',
-            'BDIM': 16
-        })
-        # we lazy-load this depending on the data types we get
-        self.fill3D_cuda = {}
         self.Ctmp = None
         
     def dot(self, A, B, out=None):
@@ -64,6 +58,18 @@ class ArrayUtilsKernel:
         
         return out
 
+    def norm2(self, A, out=None):
+        return self.dot(A, A, out)
+
+class TransposeKernel:
+
+    def __init__(self, queue=None):
+        self.queue = queue
+        self.transpose_cuda = load_kernel("transpose", {
+            'DTYPE': 'int',
+            'BDIM': 16
+        })
+
     def transpose(self, input, output):
         # only for int at the moment (addr array), and 2D (reshape pls)
         if len(input.shape) != 2:
@@ -84,8 +90,14 @@ class ArrayUtilsKernel:
         self.transpose_cuda(input, output, np.int32(width), np.int32(height),
             block=blk, grid=grd, stream=self.queue)
 
-    def norm2(self, A, out=None):
-        return self.dot(A, A, out)
+
+
+class CropPadKernel:
+
+    def __init__(self, queue=None):
+        self.queue = queue
+        # we lazy-load this depending on the data types we get
+        self.fill3D_cuda = {}
 
     def fill3D(self, A, B, offset=[0, 0, 0]):
         """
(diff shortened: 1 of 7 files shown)


bjoernenders:
You're right, these test needs to be expanded to larger sizes for unity testing. Didn't make to much sense for the references though.

ptypy/accelerate/cuda_pycuda/array_utils.py
```diff
@@ -85,6 +87,100 @@ def transpose(self, input, output):
     def norm2(self, A, out=None):
         return self.dot(A, A, out)
 
+    def fill3D(self, A, B, offset=[0, 0, 0]):
```
bjoernenders, line 102: I didn't see this before. There is no need to make this part of a "Kernel", or is there? It doesn't require a state or needs to allocate resources. We should keep those as simple functions, just like the blueprint in `base`

ptypy/accelerate/cuda_pycuda/kernels.py
```diff
@@ -3,6 +3,7 @@
 from pycuda import gpuarray
 from ptypy.utils.verbose import log, logger
 from . import load_kernel
+from .array_utils import ArrayUtilsKernel
```
bjoernenders: It should not be a Kernel

bjoernenders (changes requested):

ptypy/accelerate/cuda_pycuda/array_utils.py
xcelerit-team: The state they keep is the compiled kernel code. Without the class, you’d need to recompile on every run.

daurer: Maybe we could write the compiled kernel code into a global variable instead (and get rid of the python kernel)?

bjoernenders: Well but it doesn't make sense to compile all the other functions if you only need one either. A Kernel class only makes if the methods are somehow joined to form a greater something, like a workflow. Maybe we need to rip this class apart. Open for suggestions, but a per-function/method cache for compiled code seems appropriate.

bjoernenders: We could also use the "`__call__`" method of classes to make them appear like functions

xcelerit-team: I agree with that in general. The fill3D method however performs lazy compilation on first use and caches in the object. A strategy like this could be used globally (there are other kernel classes in ptypy where the same applies). Some kind of kernel cache. 

But I guess the question is if now is the right time to do such refactorings?

bjoernenders: the `dot` product and the reductions belong together as they depend on each, the `transpose` has nothing to do in there (it's even a different dtype here), but I didn't bother complaining last time. And that you attached these other unrelated methods is a choice and not "refactoring". However, in lieu of a strategy, we can just go ahead and make a class for each collection of codependent methods.

xcelerit-team: All the CUDA kernels are part of such kernel classes. I simply continued with that and chose what appeared to be the most suitable choice. I am open to suggestions and can move this into a kernel class with call operator for example. 

With “refactoring” I meant the rest of the GPU kernels to follow the same pattern.

bjoernenders: Oh ok, I read you wrong. No I don't mean to refactor all Kernel classes because some share data types or memory buffers and the individual methods/kernels don't really mean anything outside the class construct or the methods are codependent or called recursively.

For isolated functions (like most of those that are in base/array_utils) we should however strive to make them separate objects as much as possible. I'll make a suggestion here in this branch

bjoernenders (approved):

daurer (approved):

PR #305, opened 2021-03-12
WIP: Upsampling/Downsampling kernels

daurer:


ptypy/accelerate/base/engines/ML_serial.py
```diff
@@ -368,7 +388,13 @@ def new_grad(self):
                     f["aux"] = aux
                     f["addr"] = addr
 
-            GDK.make_model(aux, addr)
+            if kern.resample:
+                aux_tmp1 = np.abs(aux)**2
+                au.resample(aux_tmp2, aux_tmp1)
+                au.resample(aux_tmp1, aux_tmp2)
```
bjoernenders, line 468: this looks odd. Shouldn't the second resample be after make model?

bjoernenders (changes requested):
See that one comment. Otherwise go ahead.

PR #308, opened 2021-03-18
includes scratchmem size in dict key

bjoernenders:
This is the fix to get rid of the MemAlloc and MemFree. The problem was that the `scratchmem` would be made new each time `max_abs2` was applied to differently sized array, which was the feature programmed. But if we alternate between two differently sized arrays, it makes a new scratch each time. Another possible workaround would have been to use a separate `max_abs2` "Kernel" for probe and object update respectively

### DIFF
--- a/ptypy/accelerate/cuda_pycuda/array_utils.py
+++ b/ptypy/accelerate/cuda_pycuda/array_utils.py
@@ -117,9 +117,13 @@ class MaxAbs2Kernel:
 
     def max_abs2(self, X, out):
         """ Calculate max(abs(x)**2) across the final 2 dimensions"""
+        rows = np.int32(X.shape[-2])
+        cols = np.int32(X.shape[-1])
+        firstdims = np.int32(np.prod(X.shape[:-2]))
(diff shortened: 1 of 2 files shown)


daurer (approved):

xcelerit-team (changes requested):
While this will work, it will re-compile the kernel for every new array size it finds. 

I was thinking that having a different map for the scratchmem might be a better approach. Or alternatively, only reallocate if the memory is too small (but not too big). A too big array for scratch memory doesn't hurt - only a part of it will be used. The data type of the memory would still need to be considered though.

bjoernenders:
@lotzej good point. But the codes we do this for are for massive repetitions on the same structures and not to really write a library to cover all cases so I can live with the recompilation penalty. This one was quick, easy and even saves some lines of code. But feel free to push an alternative at the end of the hackathon if there are some cycles left.

daurer:
Going to merge this, we can still look into alternative solutions if there is time left at the end of the hackathon.

PR #309, opened 2021-03-22
WIP: position correction

daurer:


daurer:
@bjoernenders I have decided to completely remove position correction for ML. If we want position correction also for ML, we should implement it properly as part of the gradient update. 

If you are okay with this, I am going to merge this PR which includes a working position correction including the new grid search and covering all DM engines. I will do the DR engines in a separate PR.

PR #311, opened 2021-03-23
Remove traces of OpenCL engine

daurer:
In the OpenCL engine, there was a need to pad to object arrays. This is not necessary in the pycuda engines and can be removed.


daurer:
@bjoernenders  Can we merge this? As far as I remember this extra code is a left over from the PyOpenCL code. It's a bit annoying to not have the same object shapes in DM and DM_serial...

PR #312, opened 2021-03-26
Fix in context initialisation to raise an exception

xcelerit-team:
In case more processes are launched than GPUs, the context creation doesn't work and the cuda engines won't work anyway. So I'm proposing to raise an exception in this case.


daurer (approved):
I am okay with this, but I would make the error message a bit more verbose.

xcelerit-team:
those changes make sense to me - error message is better like this.

bjoernenders (approved):

PR #313, opened 2021-03-30
Precompile cufft during setup to avoid MPI failures and speed up execution

xcelerit-team:
We discovered that using `fork()` in an MPI process is dangerous and only supported in certain implementation and hardware setup-specific cases. The way that the `cufft` module was imported previously, compiling it as an extension module on the fly, caused crashes in the diamond cluster as soon as multiple nodes were used. Also, the cufft compilation setup takes quite long at runtime.

This pull request mitigates both these issues by:
* Building the `filtered_cufft` as an extension module during setup phase
* Puts all powers of 2 from 16 to 2048 into the binary at the moment (for the 2D image size), and throws an error if any other sizes are used (incl. non-square)
* `cufft.py` is now simply importing that pre-compiled module rather than runtime compiling
* This was tested by using `pip install --user -e .` at the top-level first, and then running the fft tests
* It was also tested in the diamond cluster - it works with MPI now

Note: In the engines, we could for example check if the sizes are powers of 2 - and if so, use cufft. Otherwise fall back to Reikna for all other cases. There won't be a startup delay with cufft anymore.


ptypy/accelerate/cuda_pycuda/cuda/filtered_fft/filtered_fft.cu
```diff
@@ -274,31 +262,55 @@ void FilteredFFTImpl<ROWS,COLUMNS,SYMMETRIC,IS_FORWARD>::setupPlan() {
...
+      throw std::runtime_error("Only equal numbers of rows and columns are supported");
+    switch (rows)
+    {
+        case 16: return new FilteredFFTImpl<16, 16, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
+        case 32: return new FilteredFFTImpl<32, 32, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
+        case 64: return new FilteredFFTImpl<64, 64, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
+        case 128: return new FilteredFFTImpl<128, 128, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
+        case 256: return new FilteredFFTImpl<256, 256, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
+        case 512: return new FilteredFFTImpl<512, 512, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
+        case 1024: return new FilteredFFTImpl<512, 512, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
```
daurer: are these typos?

ptypy/accelerate/cuda_pycuda/cuda/filtered_fft/filtered_fft.cu
```diff
@@ -274,31 +262,55 @@ void FilteredFFTImpl<ROWS,COLUMNS,SYMMETRIC,IS_FORWARD>::setupPlan() {
...
+        case 2048: return new FilteredFFTImpl<512, 512, SYMMETRIC, FORWARD>(batches, prefilt, postfilt, stream);
```
daurer: same here?

daurer (changes requested):
I think we should make the pre-compilation optional, perhaps using an environment variable? This is to consider the case the user has no GPUs, but should still be able to run ptypy as usual with no extra dependencies (e.g. pybind)

ptypy/accelerate/cuda_pycuda/cuda/filtered_fft/filtered_fft.cu
xcelerit-team: indeed - I fixed them now.

xcelerit-team:
> I think we should make the pre-compilation optional, perhaps using an environment variable? This is to consider the case the user has no GPUs, but should still be able to run ptypy as usual with no extra dependencies (e.g. pybind)

I have now made the compilation optional, using a command-line flag in setup.py. This is a quick and dirty way, which will not work if the package is installed with pip instead of running setup.py directly. A better option would be:
- put the cufft extension module as a separate python package with its own setup.py
- put an optional dependency into ptypy (`extras_require={ "cufft": ["pybind11"] }`), so that when users do `pip install ptypy` it installs it without that dependency, and if users do `pip install ptypy[cufft]` it installs the optional dependency module

Another option is to try to compile it in any case, and if it fails, retry the setup(...) call without the extension. This is done for example in `simplejson` here: https://github.com/simplejson/simplejson/blob/master/setup.py

A third option is to use an environment variable to control the setting, as sqlalchemy does for its C extensions, or detect if cuda is available on the system and enable it in this case, etc.

daurer (approved):
I have removed the command line flag to setup.py and replaced it with a try/except. If either CUDA or pybind11 is not available it triggers an exception which then prints a warning at the end of the installation process. I have also included the other alternative solutions from @lotzej for this switch as a comment in setup.py

@bjoernenders if you are okay with this current version, feel free to merge. I am going to also make changes to the choosing of the FFT library in the propagation kernel, once #310 is merged, otherwise there are going to be conflicts.

PR #322, opened 2021-04-26
bugfix: rescale size of aux when using MPI

daurer:
With running ptypy with more than a single process, and when using the new BlockScanModel, the data loader will distribute the diffraction data from a single block across all available MPI processes. So, if `frames_per_block` is 1000 and we are using 4 MPI processes, the effective data frames in a block is 250. But in our pycuda engines, we are still allocating the `aux` arrays with the full size of 1000 and we therefore significantly over-allocate `aux`. This actually makes a big and noticable difference in performance.


bjoernenders (approved):
I'm surprised we haven't noticed this earlier.

PR #327, opened 2021-05-03
gather_dict/bcast_dict can create issues with multinode MPI

daurer:
When running ptypy with multiple nodes at Diamond, I am either getting errors or extremely poor perfomance. I have identified that `parallel.gather_dict` is the cause of these issues. Replacing the current implementation of `parallel.gather_dict` with a much more simple implementation making use of `comm.gather` from mpi4py solves these issues for me. 

@bjoernenders  I would be curious to know why there is this long, quite complicated implementation in the first place?

### DIFF
--- a/ptypy/core/data.py
+++ b/ptypy/core/data.py
@@ -979,8 +979,8 @@ class PtyScan(object):
                 pos = {}
                 weights = {}
             # Distribute raw data across nodes according to indices
-            raw = parallel.bcast_dict_with_keys(raw, indices.node)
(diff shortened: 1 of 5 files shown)


bjoernenders (approved):

daurer:
The same problems occur with `bcast_dict`

bjoernenders (approved):

bjoernenders:
keep one function only (bcast_dict), remove in-place behavior. Track down all uses of in-place bcast

bjoernenders (approved):

PR #329, opened 2021-05-03
Make recording of local error map optional

daurer:
For very large datasets, saving a copy of the full map of local errors can take a long time and dramatically increase the files size. I believe this should be optional.


bjoernenders (approved):

bjoernenders:
set user level to 2 please

PR #332, opened 2021-05-14
bugfix: correctly resize the aux shape

daurer:
need to correctly expand the aux shape when crop_pad is used. Bit surprised that this was not tested during the gpu-hackathon.


daurer:
(the travis fail can be ignored, we currently don't use travis and its now deactivated)

bjoernenders (approved):

PR #335, opened 2021-05-19
Pre-determine if data is distributed

daurer:
Introduced new parameter `scaling` for the Container class.


daurer:
@bjoernenders do you have any further suggestions on this one? Otherwise, I would like to merge.

daurer:
Change variable to `distribution` with options `clone` or `scatter`

daurer:
@bjoernenders happy with this or shall we also change variable names `distributed` -> `scattered`

PR #338, opened 2021-06-07
Remove automatic import of all experiment classes

daurer:
Instead, users are required to specifically import the ptyscan class in the run script.

### DIFF
--- a/ptypy/__init__.py
+++ b/ptypy/__init__.py
@@ -72,7 +72,6 @@ del EvalDescriptor
 # Import core modules
 from . import utils
 from . import io
-from . import experiment
 from . import core
 from . import simulations
 from . import resources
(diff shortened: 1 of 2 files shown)


jcesardasilva:
hi @daurer, 

Could you please tell us if there is any example of how import a specific ptyscan class now? For example, to use `ID16AScan`. I tried, but the description says I did not provide `data.name` properly. Thank you in advance.

jcesardasilva:
Hi @daurer 

I found my mistake. I was importing from `ptypy.experiments` whereas I should have done it from `ptypy.experiment` without the `s` at the end. Now it works like this: 
```
from ptypy.experiment.ID16Anfp import ID16AScan 
```
I am sorry about the disturbance. Thank you.

PR #352, opened 2021-07-13
DM_pycuda engines don't fully clear memory

daurer:
don't merge, still in progress


daurer:
I found the problem. This is related to the cupy context that we use for the NCCL communiation. Need to check if it is possible to destroy this cupy context in engine_finalize. But also, we should probably not have NCCL enabled by default.

bjoernenders:
Maybe it's enough to clear the cupy memory pools. https://docs.cupy.dev/en/v8.3.0/reference/memory.html 
The NCCL communicator should be external to the engine ? so maybe we don't want to destroy that context with engine.finalize()

daurer:
cupy memory pools are actually empty.

``` 
mempool = cp.get_default_memory_pool()
pinned_mempool = cp.get_default_pinned_memory_pool()
print(mempool.used_bytes(), pinned_mempool.n_free_blocks())
```

The output of this executed inside engine_finalize() is 0,0.

I don't understand which object is holding the remaining device memory, but it has something to do with the NCCL communicator. We don't explicitly create a cupy context. But destroying it in engine.finalize() would be fine, because the NCCL communicator (cupy) object lives inside the pycuda engine and would be created again when a new engine is initialised (as is the pycuda context).

Don't really know where to go from here. Maybe use MPI as default communicator for now until this is resolved?

bjoernenders:
hmm this could be a pure cupy analysis, i.e. at which point does cupy fill its context with stuff. On importing NCCL? 
On importing cupy? Does cupy automatically claim a certain fraction of memory from the GPU? There is another env variable to try: `CUPY_GPU_MEMORY_LIMIT` (setting that to 0 means unlimited though, maybe use something like 5%? )

Another option would be to explicitly delete the NCCL communicator.

Maybe reverting to the pure CPU mpi version is the way to go if can't resolve this easily.

daurer:
Setting `CUPY_GPU_MEMORY_LIMIT` does not do anything. Also, deleting NCCL communicator does not help. My feeling is that the leftover memory is actually held by `pycuda` and not `cupy` but the context management is messed up (because of NCCL/cupy) and the memory does not get free'd when the pycuda context is being detached. 

I will try to reproduce this problem in a simple example outside of the DM engine. 

Yes, it probably makes sense to revert to the MPI communicator until this issue is solved.

daurer:
Here is a simple example that reproduces this problem. 

First a simple example only using pycuda:
```
import pycuda.driver as cuda
import pycuda.gpuarray as gpuarray

# Create context from pycuda
cuda.init()
ctx = cuda.Device(0).make_context()
ctx.push()

# Allocate memory on the device
a = gpuarray.zeros((1000,1000,1000), dtype='float32')

# Destroy the context from pycuda
# this should free up the memory
ctx.pop()
ctx.detach()
```

this leaves the device in a state with 4MB allocated (according to `nvidia-smi`). 

Now, also including the NCCL communicator from cupy
```
import pycuda.driver as cuda
import pycuda.gpuarray as gpuarray

# Create context from pycuda
cuda.init()
ctx = cuda.Device(0).make_context()
ctx.push()

# Initialise NCCL communicator via cupy
from cupy.cuda import nccl
id = nccl.get_unique_id()
com = nccl.NcclCommunicator(1,id,0) 

# Allocate memory on the device
a = gpuarray.zeros((1000,1000,1000), dtype='float32')

# Destroy the context from pycuda
# this should free up the memory
ctx.pop()
ctx.detach()
```

This leaves the device in a state with 4157 MB still allocated (according to `nvidia-smi`)

bjoernenders:
one last thing you could try is to delete the NCCL communicator and then garbage collect (before popping/detaching the context) and/or you could also try to pop the context an additional time.

daurer:
I have tried those things, but they make no difference.

Issue #350, opened 2021-07-01
DR as a core engine


daurer:
- replacement for the current ePIE engine
- remove ePIE from auto_load

PR #359, opened 2021-07-29
Stochastic Douglas-Rachford algorithm

daurer:
- [x] Implement a basic version of this algorithm (#350)
- [x] Renamed all related algorithms to SDR  which stands for stochastic (semi-implicit) douglas-rachford making sure it's not confused with DM
- [x] Extended the basic fourier update to include a parameter tau which is used in the SDR engine. By default, tau =1 and in that case there are no changes to the current fourier update. 
- [x] Created generalized stochastic base classes/engines (also for seiral and pycucda) with properties that allow modification of fourier,object and probe update
- [x] Created derived classes/engines for EPIE and SDR (also for serial and pycuda)
- [x] Updated CUDA kernels and tests
- [x] Merge pycuda stream features into the stochastic pycuda engine
- [x] Integrate position correction into all stochastic engines


bjoernenders (approved):
Can't find anything wrong. Good Job!

Issue #362, opened 2021-08-03
Create a Jupyter Notebook for an "interactive" experience of ptypy


bjoernenders:
Adapt ptypy in all aspects to make this work nicely.

bjoernenders:
Achieved with release 0.7.0

PR #371, opened 2021-10-13
Improve logging

daurer:
- Introduce new logging level for interactive case (e.g. jupyter notebook) #362 
- Fixed logging custom formatter (Python 2->3 leftover)
- New logging level for citation
- Added jupyter notebook using moonflower example
- New LogTimer function in utils
- Capture warnings and add to logger
- Specify logging level by strings from now on, but silently still accept integer log levels (as before)


daurer:
@bjoernenders I've made changes to the interactive logger as discussed. I have kept the `numiter` entry in the runtime dictionary, because it allows us to log the progress of the reconstruction as "Iteration # X/Y" instead of just "Iteration X". Have a look and let me know if you want any further changes, otherwise feel free to merge.

PR #373, opened 2021-10-15
Shift object and exit waves during centring probe

ptim0626:
Previously the object and exit waves are not shifted when the probe is shifted, resulting in incorrect results. After shifting the object and exit waves with the same amount of offset, it produces correct result.

### DIFF
--- a/ptypy/engines/projectional.py
+++ b/ptypy/engines/projectional.py
@@ -330,9 +330,9 @@ class _ProjectionEngine(PositionCorrectionEngine):
                         (0, c1[0], c1[1]), (0, c2[0], c2[1]))
 
                 # shift the exit waves, loop through different exit wave views
-                for k, pv in enumerate(pr_s.views):
-                    pr_s.views[k].pod.exit[:]  = u.shift_zoom(pv.pod.exit,
-                            (1.,)*2, (c1[0], c1[1]), (c2[0], c2[1]))
+                for pv in pr_s.views:
+                    pv.pod.exit = u.shift_zoom(pv.pod.exit, (1.,)*2,
+                            (c1[0], c1[1]), (c2[0], c2[1]))
 
                 log(4,'Probe recentered from %s to %s'
                             % (str(tuple(c1)), str(tuple(c2))))
--- a/ptypy/engines/projectional.py
+++ b/ptypy/engines/projectional.py
@@ -201,6 +201,9 @@ class _ProjectionEngine(PositionCorrectionEngine):
             # Overlap update
             self.overlap_update()
 
+            # Recenter the probe
+            self.center_probe()
+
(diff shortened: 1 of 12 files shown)


ptypy/engines/projectional.py
```diff
@@ -314,17 +314,25 @@ def overlap_update(self):
...
+
+                # shift the object
+                ob_s = self.ob.storages[name]
+                ob_s.data[:] = u.shift_zoom(ob_s.data, (1.,)*3,
+                        (0, c1[0], c1[1]), (0, c2[0], c2[1]))
+
+                # shift the exit waves
+                ex_s = pr_s.views[0].pod.ex_view.storage
+                ex_s.data[:] = u.shift_zoom(ex_s.data, (1.,)*3,
+                        (0, c1[0], c1[1]), (0, c2[0], c2[1]))
```
bjoernenders, line 336: Probably doesn't work with the BlockModel. We would need to account for multiple exit wave storages to be linked to one probe storage

daurer:
@ptim0626 could you also implement the modified `center_probe` feature for `ptypy/engines/stochastic.py`?

ptypy/engines/projectional.py
```diff
@@ -329,10 +329,10 @@ def center_probe(self):
...
                         (0, c1[0], c1[1]), (0, c2[0], c2[1]))
 
-                # shift the exit waves, loop through different exit wave views
-                for pv in pr_s.views:
-                    pv.pod.exit = u.shift_zoom(pv.pod.exit, (1.,)*2,
-                            (c1[0], c1[1]), (c2[0], c2[1]))
+                # shift the exit waves
+                for ex_s in self.ex.storages.values():
+                    ex_s.data[:] = u.shift_zoom(ex_s.data, (1.,)*3,
+                            (0, c1[0], c1[1]), (0, c2[0], c2[1]))
```
daurer: @ptim0626 there is a problem here when there are multiple storages (multiple scans). Since we are iterating over the probe storages already, we should only shift the exit wave storage that is associated with the probe storage. Unfortunately, the `ID` attribute of the storages differs between probe and exit, but we could potentially use the `numID` attribute to fix this, something like this:
```
for ex_s in self.ex.storages.values():
    if ex_s.numID == pr_s.numID:
        ex_s.data[:] = u.shift_zoom(ex_s.data, (1.,)*3,
                   (0, c1[0], c1[1]), (0, c2[0], c2[1]))
```

daurer: As discussed, unfortunately this approach also does not work. Best is to revert back to the previous solution iterating through all the views.

daurer:
As a final change, we should move the center_probe call out of the overlap update and place it directly in the main iterate loop between overlap and position updates...

PR #375, opened 2021-10-28
Fix how frames_per_block is defined

### LINKED ISSUE #254: pycuda: frames per block
The maximum possible nr. of frames per block should be equal to the available number of data frames. Otherwise, we are just allocating unnecessary additional memory. And maybe the default for p.frames_per_block should be less than 100000?

daurer:
In an attempt to fix #254, I made the following changes:

- New variable `max_frames_per_block` in `ScanModel` and `BlockScanModel` which dynamically defines the maximum size of a block (never groing larger than the actual size of the data)
- Access the block size in the engines via the new scan model parameter `max_frames_per_block`

This makes sure that we never allocate more GPU memory than we need and it also makes all pycuda engines compatible with the `Full` scan class irrespective of what the user might specify as `p.frames_per_block`.

### DIFF
--- a/ptypy/accelerate/base/engines/ML_serial.py
+++ b/ptypy/accelerate/base/engines/ML_serial.py
@@ -74,11 +74,7 @@ class ML_serial(ML):
             geo = scan.geometries[0]
 
             # Get info to shape buffer arrays
-            # TODO: make this part of the engine rather than scan
-            fpc = self.ptycho.frames_per_block
-
-            # When using MPI, the nr. of frames per block is smaller
-            fpc = fpc // parallel.size
+            fpc = scan.max_frames_per_block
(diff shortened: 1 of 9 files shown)


daurer:
One remaining issue is that currently the `p.frames_per_block` is defined such that it covers all MPI ranks, so effecively the block size that is used both for chunking the data and in the engines is `p.frames_per_block` divided by the number of MPI ranks. @bjoernenders it probably would make more sense to define `p.frames_per_block` per MPI rank which would make it more intuitive to the user...

daurer:
Changed it now such that the meaning of `frames_per_block` is per MPI rank.

bjoernenders (approved):

PR #382, opened 2021-12-10
[WIP] Memory management for ML_pycuda engine revised after projectional_pycuda_stream

bjoernenders:
Small tests seems to suggest it's working. Things to do:

- [x]  Check in profiler if implementation does what it is supposed to do
- [x]  Support constraint uses CPU version but should be using GPU. Solution: Copy supp contraint, but make issue for potential mixin'class
- [x]  The finalize functions differs in engine. Check for correctness.
- [x]  The position correction uses different arrays (ma & mag) and manipulates mag in place. That's incompatible and needs revision. Solution: new likelihood kernel in PosRef that uses w,I. Raise issue about derived data.
- [x]  Archive pycuda_streams and its associated memory manager.
- [x]  Clean commented text.
- [x]  Clean up memory managers.


ptypy/accelerate/base/engines/ML_serial.py
```diff
@@ -321,7 +321,7 @@ def position_update(self):
                     PCK.mangler.get_address(i, addr, mangled_addr, max_oby, max_obx)
                     PCK.build_aux(aux, mangled_addr, ob, pr)
                     aux[:] = FW(aux)
-                    PCK.log_likelihood(aux, mangled_addr, mag, ma, err_phot)
+                    PCK.log_likelihood(aux, mangled_addr, I, w, err_phot)
```
daurer: maybe should use the modified `log_likelihood_ml` kernel?

ptypy/accelerate/cuda_pycuda/engines/ML_pycuda.py
```diff
@@ -422,9 +377,11 @@ def position_update(self):
                     PCK.mangler.get_address(i, addr, mangled_addr, max_oby, max_obx)
                     PCK.build_aux(aux, mangled_addr, ob, pr)
                     PROP.fw(aux, aux)
-                    PCK.log_likelihood(aux, mangled_addr, mag, ma, err_phot)
+                    PCK.log_likelihood(aux, mangled_addr, I, w, err_phot)
```
daurer: same here, should probably be `PCK.log_likelihood_ml`?

bjoernenders: oh yes

PR #399, opened 2022-05-04
_MODE_CONV has been removed with PIL 9.1.0

daurer:
https://github.com/python-pillow/Pillow/pull/6057

To remain compatible with the latest version of PIL, I have commented the corresponding lines in `plot_utils.py` - why was this necessary in the first place? It looks like this piece of code has been there from the early days, maybe @pierrethibault remembers?


pierrethibault:
Indeed a long long time ago PIL did not support uint16 TIF for some reason, and I had found this workaround. Utterly outdated.

PR #400, opened 2022-05-20
Add L-BFGS reconstruction engine

jfowkes:
Add Limited-memory BFGS reconstruction engine as an alternative to ML.

There is room for improving the implementation coding but it will do for testing.

Caveats:
- L-BFGS does not support the object smoothing preconditioner for technical reasons (this would require the ability to invert the Gaussian filter, i.e. exactly de-blur and this is known to be impossible). However probe preconditioning (i.e. probe-object scaling) is supported and this is enough to reconstruct on the moonflower example with a lower error than ML.
- Adding support for operand `*` for `float` and PtyPy `Container` would be highly desirable as reusing the current grad container to implement this operation as is currently done in the PR is not very efficient.


jfowkes:
Thanks @ptim0626, great work on the CUDA implementation!

jfowkes:
As discussed, it is possible to avoid the inefficient copy in operations like:
```python 
#TODO: support operand * for 'float' and 'Container'
# (reusing self.ob_grad here is not efficient)
# self.ob_h -= self.alpha[i]*self.ob_y[i]
self.ob_grad << self.ob_y[i]
self.ob_grad *= self.alpha[i]
self.ob_h -= self.ob_grad
```
simply by changing the order of the `*=` and `-=` operations on the `self.ob_h` container and adding a preliminary `/=` by the scalar operation like so:
```python
# self.ob_h -= self.alpha[i]*self.ob_y[i]
self.ob_h /= self.alpha[i]
self.ob_h -= self.ob_y[i]
self.ob_h *= self.alpha[i]
```
I will implement this.

PR #406, opened 2022-10-13
Prepare public release of ptypy

bjoernenders:
- [x] Make sure the package metadata is only in one place (currently in setup.py, setup.cfg and pyproject.toml)
- [x] Optional: Simplify versioning.
- [x] Make a README for ptypy. Does that need to include the PF banner? 
- [x] Review README in general.
- [ ] Do wheel builds make sense?
- [ ] Test istalling from pypi. 
- [ ] Write up ptypy upload as dev guide.


bjoernenders:
I think we are done here.

pierrethibault:
Thanks Bjoern. That's one big ugly page being turned.

daurer (approved):

daurer (approved):

PR #408, opened 2022-11-10
[WIP] Streaming data from a live SWMR file

daurer:
The main idea with this PR is to read data from a SWMR file in blocks, the size of which can be determined by `p.frames_per_block`. To achieve this, the `check()` routine in the ptyscan class is checking if a full new block of data is available, if not it will return a WAIT signal which then skips the entire loading routine. If a full block of data is available, that block is loaded and the  loading continues until no further block is available. 

The ptycho class also has a new parameter `p.frames_before_iterate` which sets a lower limit to the nr. of frames that are required to be loaded before going into the reconstruction. Also, the total number of iterations is extended while data is still incomplete meaning that the user-specified `engine.numiter` are guaranteed to run after the full dataset has been loaded.

To summarise, this simple streaming scenario will read as much data as possible (in blocks) and then, if the minimum nr. of frames has been reached, start the reconstruction in blocks of `engine.numiter_contiguous` until all data is loaded and then run further `engine.numiter` iterations.


ptypy/experiment/hdf5_loader.py
```diff
@@ -560,8 +525,8 @@ def load_unmapped_raster_scan(self, indices):
             slow_idx, fast_idx = self.preview_indices[:, ii]
             intensity_index = slow_idx * sh[1] + fast_idx
             weights[ii], intensities[ii] = self.get_corrected_intensities(intensity_index)
-            positions[ii] = np.array([self.slow_axis[slow_idx, fast_idx] * self.p.positions.slow_multiplier,
-                                      self.fast_axis[slow_idx, fast_idx] * self.p.positions.fast_multiplier])
+            positions[ii] = np.array([self.slow_axis[slow_idx, fast_idx].squeeze() * self.p.positions.slow_multiplier,
```
daurer: This should probably be changed to 
```
slow_axis = self.slow_axis[...]
self.slow_axis = np.squeeze(slow_axis) if slow_axis.ndim > 2 else slow_axis
positions[ii] = np.array([self.slow_axis[slow_idx, fast_idx] * self.p.positions.slow_multiplier
```
to be consistent with the previous version of the code.

ptypy/experiment/hdf5_loader.py
```diff
@@ -363,51 +324,51 @@ def __init__(self, pars=None, **kwargs):
...
-            fast_axis = fast_axis[self.p.outer_index]
+            fast_axis = self.fast_axis[self.p.outer_index]
         self.fast_axis = np.squeeze(fast_axis) if fast_axis.ndim > 2 else fast_axis
         positions_fast_shape = self.fast_axis.shape
 
-        slow_axis = h5.File(self.p.positions.file, 'r')[self.p.positions.slow_key][...]
+        slow_axis = h5.File(self.p.positions.file, 'r', swmr=self._is_swmr)[self.p.positions.slow_key]
         if self._is_spectro_scan and self.p.outer_index is not None:
-            slow_axis = slow_axis[self.p.outer_index]
+            slow_axis = self.slow_axis[self.p.outer_index]
```
daurer, line 361: this should be without `self.`

daurer:
replaced by #428

PR #411, opened 2022-11-23
should use meta instead of info to get psize

daurer:


daurer:
After discussion with Bjoern, decided to fix this issue by populating `ptyscan.info` instead of `ptyscan.meta` in the HDF5Loader class.

Issue #474, opened 2023-01-28
v0.7 - multi modes broken in DM / DM_pycuda / ML_pycuda engines.


kahntm:
Hej ptypy community,

I recently upgraded to ptypy 0.7 and enjoy using the DM / ML / DM_pycuda and ML_pycuda engine with one probe and one object mode without any issues.

But I noticed an issue when using multiple probe modes together with the DM engine. While using multiple object modes works flawlessly, I encounter error messages, the moment I want to use more than one probe mode and the DM engine. 
![2_probe_modes_DM](https://user-images.githubusercontent.com/26224125/215279236-0c0adbb3-4de9-400a-be95-98b2a4992541.png)

With the ML engine, the use of multiple probe and multiple object modes works without any errors and seems to give meaningful results.

When using DM_pycuda, both the use of either multiple probe modes or multiple object modes does not result in an error, but instantly creates NANs in with the first iteration. 

When using ML_pycuda, the usage of multiple object modes works without issues. But the usage of multiple object modes throws an error:
![4_probe_modes_ML_pycuda](https://user-images.githubusercontent.com/26224125/215283071-d28a5b66-649b-4b46-a069-b62e032de5a6.png)

Any help / insights are appreciated. I hope it is not just my installation.

daurer:
Thanks @kahntm for reporting this issue, it seems like this is a bug introduced by this [commit](https://github.com/ptycho/ptypy/commit/8b4a60d61a5ba60d850971bf62329c7f741a029f). I am working on a fix.

daurer:
This has been patched with [v0.7.1](https://github.com/ptycho/ptypy/releases/tag/v0.7.1)

PR #475, opened 2023-01-29
patch to make numpy FFT C-contiguous and allow user to choose FFT type

daurer:
This resolves the issue reported in #474 that was caused by this recent [commit](https://github.com/ptycho/ptypy/commit/8b4a60d61a5ba60d850971bf62329c7f741a029f).

The underlying problem was that the Numpy FFT changes the `strides` of the ndarray and makes the target array F-contiguous which causes both in our MPI and CUDA implementations that require C-contiguous arrays. I have added `np.ascontiguousarray` to both the `fft` and `ifft` of the numpy FFT propagator. 

As a result, the numpy FFT might not be the fastest option, therefore we should allow the user to choose between the 3 available FFT types (numpy, scipy and fftw) - I added a new parameter `p.ffttype` in the scan model. This is a little inconsistent with the CUDA engines which define the FFT library in the engines. 

Also, its not quite clear what would be the best default choice for `ffttype`, currently it is set to "numpy".

### DIFF
--- a/ptypy/core/geometry.py
+++ b/ptypy/core/geometry.py
@@ -107,6 +107,13 @@ class Geo(Base):
     doc = Either "farfield" or "nearfield"
     userlevel = 1
 
+    [ffttype]
+    type = str
+    default = numpy
+    help = FFT library
+    doc = Either "numpy", "scipy" or "fftw"
+    userlevel = 1
+
     [shape]
     type = int, tuple
     default = 256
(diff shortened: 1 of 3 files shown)


daurer:
Benchmarking the speed of the different FFT libraries:

**Nearfield**
`data = np.random.random((20,2048,2048)) + 1j * np.random.random((20,2048,2048))`

`%timeit out = prop_numpy.fft(data)`
5.64 s ± 21 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

`%timeit out = prop_scipy.fft(data)`
3.62 s ± 181 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

`%timeit out = prop_fftw.fft(data)`
1.49 s ± 15.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

**Farfield**
`data = np.random.random((100,512,512)) + 1j * np.random.random((100,512,512))`

`%timeit out = prop_numpy.fft(data)`
1.22 s ± 6.13 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

`%timeit out = prop_scipy.fft(data)`
738 ms ± 1.37 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

`%timeit out = prop_fftw.fft(data)`
310 ms ± 5.14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

daurer:
Based on the quick benchmark above, the winner is `fftw` but I would vote for making `scipy` the default choice considering that the FFTW library is not part of the core dependencies. @bjoernenders what do you think?

PR #484, opened 2023-03-03
Added Dockerfile and new CLI script

daurer:
Aiming to make PtyPy ready to run in cloud environments

Splitting this into multiple PRs


ptypy/accelerate/cuda_cupy/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# Set environment variables
+ENV CUDA_PATH=/usr/local/cuda \
+    LD_LIBRARY_PATH=$CUDA_PATH/lib64:$LD_LIBRARY_PATH
+
+# Clone PtyPy
+RUN git clone https://github.com/ptycho/ptypy.git
+
+# Conda env for PTYPY CUDA/PYCUDA backend
+RUN conda env create -f ptypy/ptypy/accelerate/cuda_cupy/dependencies.yml \
+    && conda install -n ptypy_cupy pytest
```
garryod: When working in a container it is conventional to update the base environment (as it should be the only one)

```suggestion
RUN conda env update --name base --file ptypy/ptypy/accelerate/cuda_cupy/dependencies.yml \
    && conda install --name base pytest
```

ptypy/accelerate/cuda_cupy/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    && conda install -y python=3 conda-build\
+    && conda update -y conda \
+    && conda clean --all --yes
+
+# Set environment variables
+ENV CUDA_PATH=/usr/local/cuda \
+    LD_LIBRARY_PATH=$CUDA_PATH/lib64:$LD_LIBRARY_PATH
+
+# Clone PtyPy
+RUN git clone https://github.com/ptycho/ptypy.git
```
garryod: It would likely be better to `COPY` here

```suggestion
COPY ./ ./
```

ptypy/accelerate/cuda_cupy/Dockerfile
```diff
@@ -0,0 +1,33 @@
+# This includes cuda devtools
+FROM nvidia/cuda:12.0.1-devel-centos7
```
garryod: Is there a particular reason we're using `centos` for this container? Afaik it's less well supported by Nvidia

garryod (changes requested):
Looks like a good start, tough I've got a few questions & suggestions:
- It is conventional to have a dockerfile at the root of the repo, is there a reason you wouldn't want to do this?
- It should be possible to combine these two `Dockerfile`s and still produce two images by using a build arg - e.g. `ARG CUDA_LIB=cupy`, which could then be specified during `docker build`

garryod:
It may also be worth building & publishing the container to the GitHub Container Registry (`ghcr.io`) on releases. I've got a reference workflow for this in the[ `bookshelf-template`](https://github.com/DiamondLightSource/bookshelf-template/blob/9231d65949c5703f037e17e5db36ef9770068f5e/.github/workflows/containers.yml)

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY ./dependencies_core.yml ./dependencies.yml
+RUN sed -i "s/python=3.9/python=${PYTHON_VERSION}/" dependencies.yml
+RUN conda env update -n base -f dependencies.yml && conda install -y -n base pytest
+
+# Pull from existing image with ptypy dependencies and set up testing
+FROM localhost/ptypy_${PLATFORM}_py${PYTHON_VERSION}_devel:latest as runtime
+ENV WORKDIR=/ptypy
+WORKDIR ${WORKDIR}
+COPY . ${WORKDIR}
+RUN cd ${WORKDIR} && pip install .
```
garryod: Your `WORKDIR` is the context in which `RUN`/`COPY` etc. operates, so you can forgo specifying it explicitly
```suggestion
COPY ./ ./
RUN pip install .
```

.github/workflows/containers.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Build Image and push to local registry
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          push: true
+          tags: localhost:5000/ptypy/${{ matrix.platform }}_py${{ matrix.python-version }}_dev:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          target: ${{ env.TARGET }}
+          file: ./Dockerfile
```
garryod: This is the default and can be elided
```suggestion
```

.github/workflows/containers.yml
```diff
@@ -0,0 +1,63 @@
...
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action/@v2
```
garryod: Can get rid of QEMU unless we want to perform a multi-architecture build. For this you would need to specify `platforms` in `build-push-action`

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Pull from miniconda image and install core dependencies
+FROM continuumio/miniconda3 as core
+
+ARG PYTHON_VERSION=3.9
+COPY ./dependencies_core.yml ./dependencies.yml
+RUN sed -i "s/python=3.9/python=${PYTHON_VERSION}/" dependencies.yml
+RUN conda env update -n base -f dependencies.yml && conda install -y -n base pytest
+
+# Pull from existing image with ptypy dependencies and set up testing
+FROM localhost/ptypy_${PLATFORM}_py${PYTHON_VERSION}_devel:latest as runtime
```
garryod: Pulling from a local registry is generally discouraged as the dependency is never properly captured, if you were to change this to using accelerate or core as appropriate you could have the same effect with caching

daurer: makes sense, have changed this to pull from core,pycuda or cupy

.github/workflows/containers.yml
daurer: right, removed for now

PR #485, opened 2023-03-17
Add L-BFGS reconstruction engine

jfowkes:
Resurrecting #400 as GitHub broke the original PR.

Add Limited-memory BFGS reconstruction engine to as an alternative to ML.

Caveats:

- L-BFGS does not support the object smoothing preconditioner for technical reasons (this would require the ability to invert the Gaussian filter, i.e. exactly de-blur and this is known to be impossible). However probe preconditioning (i.e. probe-object scaling) is supported and this is enough to reconstruct on the moonflower example with a lower error than ML.


jfowkes:
@daurer this L-BFGS pull request is now ready for review and potential merge.

daurer:
Needs a CuPy engine before it can be merged. @ptim0626 maybe something for you?

ptim0626:
A quick comment, the tolerance for LBFGS's tests is increased as it is very likely the discrepancy is a result of numerical instability instead of the problem of implementation.

jfowkes:
@daurer this PR will soon need to be updated to use the wavefield preconditioner, I'm happy to do this.

jfowkes:
@daurer I've now added the wavefield preconditioner to LBFGS with tests and templates. Unfortunately pytest has stopped picking up the LBFGS tests in the CI, I assume this is the same issue that we had with the ML engine tests not being picked up?

daurer:
> @daurer I've now added the wavefield preconditioner to LBFGS with tests and templates. Unfortunately pytest has stopped picking up the LBFGS tests in the CI, I assume this is the same issue that we had with the ML engine tests not being picked up?

just saw this comment now. I think the tests should be name `<something>_test.py`

jfowkes:
> just saw this comment now. I think the tests should be name `<something>_test.py`

Done. All tests now running (aside from the serial wavefield caveat as in #607) and passing.

jfowkes:
@daurer there are small differences in the object due to slightly different optimization paths between `LBFGS` and `LBFGS_serial` but nothing that I would be concerned about. Visually the reconstructions look exactly the same so I have simply upped the tolerance for the failing `test_LBFGS_serial_floating` from 0.1 to 0.2 and the tests pass.
Errors

daurer (approved):

Issue #490, opened 2023-07-15
Switching between engines


kahntm:
It is quite common to run X iterations of for example DM or RAAR followed by Y iterations of ML. For a test I tried to follow that up by more engines. For example: DM + ML + DM + ML ... 
And I noticed that, while I can switch as often as I want between DM_pycuda and RAAR_pycuda, and go once to ML_pycuda, the 2nd time I want to run ML_pycuda, ptypy always crashes.

When ptypy tries to init the ML engine a 2nd time this is printed by every worker
`  * fft_lib              : reikna
================================================================================
Traceback (most recent call last):
  File "/asap3/petra3/gpfs/p06/2023/data/11016011/processed/macros/rec_single_GPU_test.py", line 254, in <module>
    P = Ptycho(p,level=5)
        ^^^^^^^^^^^^^^^^^
  File "...ptypy/build/lib/ptypy/core/ptycho.py", line 384, in __init__
    self.run()
  File ".../ptypy/build/lib/ptypy/core/ptycho.py", line 763, in run
    self.run(engine=engine)
  File ".../ptypy/build/lib/ptypy/core/ptycho.py", line 649, in run
    engine.initialize()
  File ".../ptypy/build/lib/ptypy/engines/base.py", line 153, in initialize
    self.engine_initialize()
  File ".../ptypy/build/lib/ptypy/accelerate/cuda_pycuda/engines/ML_pycuda.py", line 101, in engine_initialize
    super().engine_initialize()
  File ".../ptypy/build/lib/ptypy/accelerate/base/engines/ML_serial.py", line 48, in engine_initialize
    super(ML_serial, self).engine_initialize()
  File ".../ptypy/build/lib/ptypy/engines/ML.py", line 160, in engine_initialize
    self.ob_grad = self.ob.copy(self.ob.ID + '_grad', fill=0.)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../ptypy/build/lib/ptypy/core/classes.py", line 1740, in copy
    new_cont = type(self)(self.owner,
               ^^^^^^^^^^^^^^^^^^^^^^
  File ".../ptypy/build/lib/ptypy/core/classes.py", line 1574, in __init__
    super(Container, self).__init__(owner, ID)
  File "...ptypy/build/lib/ptypy/core/classes.py", line 121, in __init__
    owner._new_ptypy_object(obj=self)
[...]

PR #491, opened 2023-07-19
Properly clean up accelerated ML engines to allow chaining

daurer:
`ML_serial` should call `ML.engine_finalize` in order to properly clean up containers to allow accelerated ML engines to be chained.


daurer:
this fixes #490

PR #493, opened 2023-07-19
Count usable CPU number

ptim0626:
This PR fixes a small issue when counting the number of CPU. The usable number of CPUs should be used, instead of counting all the CPUs available, this is especially important for using this in HPC system, otherwise it creates more instances of multiprocessing than the number of usable CPUs, resulting in slow-down. The way of counting usable number of CPUs is stated in the [Python documentation](https://docs.python.org/3/library/os.html#os.cpu_count).


daurer (approved):
Thanks for the fix TIm! This only affects the Hdf5 loader which uses multiprocessing, going to merge.

PR #500, opened 2023-08-23
Starting point for implementation of multislice ePIE

daurer:
- provides extra storage for slices
- shows how use nearfield propagator inside the engine
- tried to implement multislice ePIE with existing fourier/object/probe functions


daurer:
Next steps

- Double check the implementation making sure it does what its supposed to do
- Save additional containers (objects, probes, exits) into an output file, currently they are not saved into .ptyr file
- Consider modes

yiranlus:
I modified the update loop and now it works on two-slice moon flow sample. I also add the function to save the slice information into another file.

In the .ptyr file, I save the object function as the product of all the slices.

daurer:
Thanks very much @yiranlus for those modifications. Do you have any actual multislice data (real or simulated) that you could further test this code with? We could also adapt the moonflower simulation code to create some fake data that actually originates from multiple slices...

@kahntm maybe you could also try it out with some of your multislice datasets and make modifications to the code if needed?

kahntm:
> @kahntm maybe you could also try it out with some of your multislice datasets and make modifications to the code if needed?

I checked the implementation with some [old published data](https://zenodo.org/records/3978334) and was able to recreate my result from back then. The implementation of multi-slice ePIE works both when used with one or two slices.

test_ePIE_multislice

kahntm:
Switched from the one pod per view assumption to multiple pods per view.
Required various for loops here and there. But essentially the same code.
Test on actual data looks like the modes are reasonable and slice separation still works.

ptypy/custom/ePIE_multislice.py
```diff
@@ -0,0 +1,181 @@
...
+    def multislice_update(self, view):
+        """
+        Does multislice ePIE
+        based on https://doi.org/10.1364/JOSAA.29.001606
+        """
+
+        for name, pod in view.pods.items():
+            # Forward multislice, calculate exit waves
+            for i in range(self.p.number_of_slices-1):
+                # exit wave for this slice
```
daurer: Thanks @kahntm for expanding this code to use modes! Maybe switch these loops around? Shouldn't make any difference to the calculations but its more consistent with the backwards update further below.

kahntm: Possible, but it would result in a 2nd loop over all pods in the view to update the the exit wave in the last slice and update the pods probe and object
image

Alternatively the loop would be extended one index further and and if i==self.p.number_of_slices-1 would need to be added within the inner loop

daurer: I think having a second loop over the pods should be fine. I doubt it would make much of a difference performance wise...

ptypy/custom/ThreePIE.py
```diff
@@ -34,15 +34,15 @@ class ThreePIE(stochastic.EPIE):
 
     [slice_thickness]
     default = 1e-6
-    type = float
+    type = float, list, tuple
```
yiranlus: `slice_thickness` can be a single value or a list of values. When a single value is used, all the slices are assumed to have the same thickness.

kahntm: Many thanks

PR #504, opened 2023-08-25
WIP GPU implementation of FFT-based smoothing in ML

pierrethibault:
Implemented only in cuda_pycuda for now, and does not support stacks of objects.


ptim0626:
Hello, I temporarily added a dummy `_get_smooth_gradient_fft` in the serial version to avoid calling the old GSK while using `ML_pycuda`.

jcesardasilva:
> Hello, I temporarily added a dummy `_get_smooth_gradient_fft` in the serial version to avoid calling the old GSK while using `ML_pycuda`.

Ok, thank you @ptim0626. May I kindly ask for your recommendations on the best course of action to pursue? Additionally, are we looking to explore a more long-term solution to the matter at hand? What are the options?

ptim0626:
Hi @jcesardasilva, the important thing is making sure `ML_pycuda` here uses the FFT-based smoothing in every smoothing operation, the fix here ensures this.

As with how to proceed, I would suggest first and foremost this is working in some testing dataset, and get the corresponding kernel in cuda_cupy. What do you think? @pierrethibault @bjoernenders @daurer

daurer:
The current solution seems to fix the problem that was reported when using the smoothing preconditioner in ML_pycuda. But more work is need before this can be merged. 

- should also be implemented in ML_cupy
- should we provide both convolution-base and fft-based version and let user decide or silently replace with FFT version in pycuda/cupy engines?
- do we need a similar change in the object smoothing regulariser in DM etc.?

pierrethibault:
I would just replace the convolution without adding an option, and use this in DM object smoothing also.

daurer:
FFT-based smoothing is now implemented as a kernel for both PyCUDA and CuPy. 
The fft-based smoothing regulariser is available in both ML_pycuda and ML_cupy by changing the new parameter `smooth_gradient_method` from "convolution" (default) to "fft".

FFT-based smoothing for other use cases, i.e. object smoothing in DM is still outstanding, see #546

daurer (approved):

PR #506, opened 2023-08-29
[WIP] Make it easier to do parameter sweeps

daurer:
without needing to load the data multiple times, this is to address #465


daurer:
avoid save/load logic, use copy_state and restore state instead. Define new iternal variable with state. Use container copies.

daurer:
@bjoernenders are you happy with this implementation of copying and restoring the state of reconstructions.

bjoernenders (approved):

PR #509, opened 2023-10-13
if refined positions are saved, they are also saved to dump files

kahntm:
When position correction is used within an engine, up to now, the refined positions were only saved in the last/reconstruction file written by that engine. As it can be interesting to track the refinement process of the scan positions over the iterations, it would be nice to also have the possibility to save the refined positions to the dump files.

The changes to ptypy/core/ptycho.py accomplish just this. Once the record_positions parameter is set, dump files also get filled with the positions. By default this parameter is still set to False. So it is a conscious decision by the user to have the positions saved.

### DIFF
--- a/ptypy/core/ptycho.py
+++ b/ptypy/core/ptycho.py
@@ -989,6 +989,11 @@ class Ptycho(Base):
                 if len(self.runtime.iter_info) > 0:
                     dump.runtime.iter_info = [self.runtime.iter_info[-1]]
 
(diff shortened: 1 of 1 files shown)


daurer:
@kahntm thanks for putting this PR, could you please rebase your commit on top of `dev` and then also change the target branch in the PR to `dev`
Thanks

kahntm:
rebased the commit and changed the target in the PR

Issue #511, opened 2023-11-16
Recontructed object larger than 50Mb issue


ltang320:
In our reconstruction, we want a large object under 10nm pixel size. However, the object size will be larger than 50Mb. In this case, it will raise the error that arrays larger than 50M are unsupported. Is there any way to fix it?
@daurer @bjoernenders 

`    
  raise RuntimeError('Arrays larger than %dM not supported. You '
  RuntimeError: Arrays larger than 50M not supported. You requested 193.99M pixels.
      raise RuntimeError('Arrays larger than %dM not supported. You '
  RuntimeError: Arrays larger than 50M not supported. You requested 193.99M pixels.
      raise RuntimeError('Arrays larger than %dM not supported. You '
  RuntimeError: Arrays larger than 50M not supported. You requested 193.99M pixels.
      raise RuntimeError('Arrays larger than %dM not supported. You '
  RuntimeError: Arrays larger than 50M not supported. You requested 193.99M pixels.

  Primary job  terminated normally, but 1 process returned
  a non-zero exit code. Per user-direction, the job has been aborted.

  mpirun detected that one or more processes exited with non-zero status, thus causing
  the job to be terminated. The first process to do so was:
  
    Process name: [[34300,1],3]
    Exit code:    1
`

daurer:
Hi,
Sure, we can increase that limit if needed. However, I am wondering if you are indeed running into this limit or if perhaps there might be something else wrong in your geometry. 

The error message suggests that ptypy is trying to create an object with 194 Million pixels, which equals to an array of 13928388x13928388 pixels. At 10 nm pixel size, this would give you a field of view of 139 mm - this seems rather large, does it make sense?

Again, happy to make a change to extend the limit but I would double check your other parameters, i.e. detector pixel size, photon energy and scan positions and make sure they are provided with the correct units.

ltang320:
> Hi, Sure, we can increase that limit if needed. However, I am wondering if you are indeed running into this limit or if perhaps there might be something else wrong in your geometry.
> 
> The error message suggests that ptypy is trying to create an object with 194 Million pixels, which equals to an array of 13928388x13928388 pixels. At 10 nm pixel size, this would give you a field of view of 139 mm - this seems rather large, does it make sense?
> 
> Again, happy to make a change to extend the limit but I would double check your other parameters, i.e. detector pixel size, photon energy and scan positions and make sure they are provided with the correct units.

[...]

PR #513, opened 2023-11-21
Raise megapixel limit to 500

daurer:
This is a quick change in response to #511 - we should probably make this into a parameter or define dynamically based on available memory.


daurer:
increase default to 100 and change from error to warning

PR #518, opened 2023-12-04
Some precision improvements for serial kernels 

ptim0626:
This PR contains several fixes that improve the precision in some serial kernels to match those in the normal implementation. 

- the same epsilon is used to avoid division by zero, as in [here](https://github.com/ptycho/ptypy/blob/2ab0bb42db962a32aa839aead5fedd7cfa75bf1a/ptypy/engines/utils.py#L221-L227)
- use PtyPy's `abs2` to calculate `af`
- ~~do not re-calculate `af` by `fdev + mag`, but storing it~~
- use the exact same formula to propagate as in [here](https://github.com/ptycho/ptypy/blob/2ab0bb42db962a32aa839aead5fedd7cfa75bf1a/ptypy/engines/utils.py#L168)

The operations above are mathematically the same before and after the changes, but associative property does not generally hold for floating-point arithmetic, so the above changes ensure the implementation in serial kernels can get the same precision with the normal implementation.  

EDIT: Memory is probably more important than precision here, so we don't use additional storage for `af`.

### DIFF
--- a/ptypy/accelerate/base/kernels.py
+++ b/ptypy/accelerate/base/kernels.py
@@ -443,11 +443,12 @@ class AuxiliaryWaveKernel(BaseKernel):
         aux = b_aux[:maxz * nmodes]
         flat_addr = addr.reshape(maxz * nmodes, sh[2], sh[3])
         rows, cols = ex.shape[-2:]
+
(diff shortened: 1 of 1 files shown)


ptim0626:
Close for now as each of the suggestions needs to be further investigated, will be addressed separately.

Issue #497, opened 2023-08-23
Smoothing kernel for ML_pycuda croaks with memory issues for values of 200


bjoernenders:
This requires some investigation into the smoothing kernel. For certain sizes, shared memory doesn't work anymore.
MemoryError("Cannot run kernel in shared memory")
We need to figure out the limit and switch to a FFT method when the smoothing kernel gets too large.

@jcesardasilva

jcesardasilva:
Thank you, @bjoernenders 

Just adding some bits of information to understand the issue, here are the messages I get when the error happens: 

<img width="976" alt="image" src="https://github.com/ptycho/ptypy/assets/11873140/85ee80fd-5944-4c52-ba85-8ed11390c0c6">

ptim0626:
The shared memory size limit in the smoothing kernel is imposed [here](https://github.com/ptycho/ptypy/blob/4cce5ae6a57ef8ca923ace0a902e504f6430fb5e/ptypy/accelerate/cuda_pycuda/array_utils.py#L356), which is 48kb.  The 48kb, according to Nvidia, is for compatibility with different architecture (see for example [here](https://docs.nvidia.com/cuda/volta-tuning-guide/index.html#unified-shared-memory-l1-texture-cache)).

The 48kb, in my opinion, is quite conservative, it is for device with compute capability 2.x. P100 is 64kb; V100 is 96kb; H100 is 228kb! However, to enable a shared memory limit larger than 48kb, this requires an explicit [opt-in](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#shared-memory-7-x). It seems CuPy provides [a way](https://github.com/cupy/cupy/pull/2369) to do it, but not pycuda.

While gathering the numbers, according to [Nvidia](https://docs.nvidia.com/cuda/ada-tuning-guide/index.html#unified-shared-memory-l1-texture-cache), CUDA reserves 1kb per thread block for shared memory. So for guarding this, the limit can be set as something like `(<SHM_LIMIT>-1024) // 2`. 

EDIT: Sorry for my oversight! pycuda actually exposes the shared memory [opt-in](https://documen.tician.de/pycuda/driver.html#pycuda.driver.function_attribute) as well.

jcesardasilva:
Hello everyone, 

[...]

PR #519, opened 2023-12-05
Maximise the usage of dynamic shared memory for GaussianSmoothingKernel

ptim0626:
This PR maximises the usage of available dynamic shared memory for a particular GPU device by explicitly opt-in for the `GaussianSmoothingKernel`. Both pycuda and CuPy version are modified, and a flag is added to the respective `load_kernel` function to support this behaviour. The modification of `load_kernel` allows future kernels to utilise the full shared memory capacity, should the need arises.

The benefit of this is that `GaussianSmoothingKernel` is able to support a larger `smooth_gradient`, as the smoothing radius is limited by the 'halo' in the CUDA kernel which uses dynamic shared memory. 

However, **this does not completely solve #497 as it still limits `smooth_gradient`, but it does allow a larger value if your GPU has a shared memory capacity greater than 48 kB**. Smoothing in the reciprocal space #504 will remove this limit.

### DIFF
--- a/ptypy/accelerate/cuda_cupy/__init__.py
+++ b/ptypy/accelerate/cuda_cupy/__init__.py
@@ -31,7 +31,9 @@ def get_context(new_queue=False):
     return queue
 
 
-def load_kernel(name, subs={}, file=None, options=None):
+def load_kernel(name, subs={}, file=None, options=None, use_max_shm_optin=False):
+
+    global device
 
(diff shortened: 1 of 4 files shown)


daurer:
After internal discussion, we decided to not modify the setting for using shared memory but rather use fft-based smoothing across all engines #504

PR #533, opened 2024-02-02
Remove new_context argument in WASP_pycuda engine

ptim0626:
This PR removes the argument `new_context` in the WASP pycuda engine because of #520.


ptim0626:
@bjoernenders @daurer quick fix for the WASP_pycuda engine

PR #549, opened 2024-03-16
Hotfix: DM_cupy is getting stuck if nccl get_unique_id throws error

daurer:
We need to wrap nccl.get_unique_id() in a try/except to make sure we don't get stuck at the `parallel.bcast`. If NCCL is not configured we are still escaping the NCCL communicator class at `self.com = nccl.NcclCommunicator()`.

### DIFF
--- a/ptypy/accelerate/cuda_cupy/multi_gpu.py
+++ b/ptypy/accelerate/cuda_cupy/multi_gpu.py
@@ -98,7 +98,10 @@ class MultiGpuCommunicatorNccl(MultiGpuCommunicatorBase):
         # get a unique identifier for the NCCL communicator and
         # broadcast it to all MPI processes (assuming one device per process)
         if self.rank == 0:
-            self.id = nccl.get_unique_id()
+            try:
+                self.id = nccl.get_unique_id()
+            except:
+                self.id = None
         else:
             self.id = None
 


bjoernenders (commented):
Hi Benedikt, I'm not quite sure what this solves. Shouldn't this fail earlier if nccl isn't configured? Like when we initialize the communicator.

daurer:
Hi Björn,

With the current code it does not fail earlier. With the original pycuda multi gpu code, we used mpi comm by default. With cupy, this was changed to make nccl the default but fall back to mpi comm if there is any error. In other words have_nccl = True always.

So the first place when it fails is in get_unique_id() but because it fails only on rank 0 and not the other ranks we get stuck at the bcast. 

The issue has been reported initially by @kahntm who noticed that DM_cupy did not work and I could then also replicate at NERSC. On my local test system I didn't catch the issue because nccl was configured. 

Now, whether it should have failed earlier is a different question. Any suggestions how we can make it fail earlier? Or switch to mpi comm by default?

daurer:
> Shouldn't this fail earlier if nccl isn't configured? Like when we initialize the communicator.

This bug fix is in the init of the communicator.

bjoernenders (approved):

bjoernenders:
I think we should some point get to a more elegant solution that this, but I don't have one on hand. Let's just fix it with this for now.

PR #555, opened 2024-06-27
Deprecation of the top-level cmap registration and access functions in `mpl.cm` in `plot_utils.py` from matplotlib 3.9.0

jcesardasilva:
Hi everyone, 

It has been a while since Matplotlib has been informing us about the deprecation of the top-level camp registration and access functions in `mpl.cm`, but now, in version Matplotlib 3.9.0, I noticed that the functions have definitely changed and it is no longer a simple deprecation warning. More details here: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.9.0.html#top-level-cmap-registration-and-access-functions-in-mpl-cm

Since `ptypy` uses them often in `plot_utils.py`, I don't know what you plan to do. 
I made the changes that worked for me, but I will let you check them. If I understood it correctly, those changes should be compatible with Matplotlib>3.6.0 from 2022, so if we implement them, `ptypy` should not be used with older versions of Matplotlib. 

Please let me know what you decide. You can decline this PR if you want.

### DIFF
--- a/ptypy/utils/plot_utils.py
+++ b/ptypy/utils/plot_utils.py
@@ -345,7 +345,7 @@ def imsave(a, filename=None, vmin=None, vmax=None, cmap=None):
     uses a matplotlib colormap with name 'gray'
     """
     if str(cmap) == cmap:
-        cmap = mpl.cm.get_cmap(cmap)
+        cmap = mpl.colormaps.get_cmap(cmap)
 
     if a.dtype.kind == 'c':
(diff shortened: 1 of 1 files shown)


jcesardasilva:
Sorry, I did it on the `master` branch and should have done it in the `dev` branch, correct? I noticed that `dev` branch was some commits behind the `master`, but I did not want to make the `rebase` of it, which I Iet for the superdeveloppers @daurer and @bjoernenders

Issue #148, opened 2018-04-23
Installation requirements.txt


benglvr:
Suggestion: Create requirements.txt file (pip freeze > requirements.txt) which can be installed with pip install -r requirements.txt for easy installation.

For debian 8 / python 2.7.14 in a fresh pyenv, the following is installed (using release_candidate branch)
```
backports.functools-lru-cache==1.5
cycler==0.10.0
h5py==2.7.1
kiwisolver==1.0.1
matplotlib==2.2.2
mpi4py==3.0.0
numpy==1.14.2
Pillow==5.1.0
pyFFTW==0.10.4
pyparsing==2.2.0
python-dateutil==2.7.2
pytz==2018.4
pyzmq==17.0.0
scipy==1.0.1
six==1.11.0
subprocess32==3.2.7
```

Tests: 3 failed, 133 passed (?)
https://gist.github.com/benglvr/4e8fbebc778f8697292487542cf019e0
minimal_prep_and_run.py runs and generates an image as expected

Additionally the following system packages were required for some of these pip extensions to compile (and to run ptypy under mpiexec) - should be added to documentation
sudo apt-get install openmpi-bin
sudo apt-get install mpich2
sudo apt-get install libopenmpi-dev
sudo apt-get install libfftw3-dev
There may be other requirements that I already have installed.

aaron-parsons:
Hi Ben,

Sorry it took me so long to get back to you! I was involved in some projects.

Currently the way we manage our dependencies is via conda. You'll notice an environment.yml file in the repo which you can activate using conda env create --file environment.yml. The rest of the install commands are inside .travis.yml file which is tested by travis every time someone pushes to their remote.

This means that our versions are always up to date with the latest versions available in conda and we spot any bugs ahead of time.

We are currently revising the dependencies to make sure only core dependencies are required, with additional optional dependencies supplying additional functionality.

We will change this for 0.4 to rely on conda recipes instead which are much nicer.

Have you tried the environment.yml? Did you find any bugs?

benglvr:
Hi Aaron,

No worries about the delay. I'll have a go on Monday installing this under conda in a fresh virtual machine and get back to you.

Regards

Nordicus:
Trying to get started as a new user, I think this is possibly the correct open issue to continue.
The documentation is currently somewhat unclear on how to perform a clean installation.

[...]

PR #570, opened 2024-08-14
Add requirements.txt

TsynkPavel:
Building by means of conda is not very friendly for new users. It is suggested to add familiar to python developers requirements.txt.

Specific version of matplotlib was chosen specially, because when installing the latest version, I get the error `AttributeError: module 'matplotlib.cm' has no attribute 'register_cmap'`. With version 3.7.3 there are no such problems. 

issue: #148

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,10 @@
+numpy
+matplotlib==3.7.3
+h5py
+pyzmq
+mpi4py
+pillow
+pyfftw
+pyyaml
+scipy
+cupy


KrashAnarhia:
I think  with venv (without conda) will working is better. Why are u use conda env?

daurer:
@TsynkPavel Thanks for your sharing your concerns. There is no need for a `requirements.txt` file since the dependencies are already specified in the [pyproject.toml](https://github.com/ptycho/ptypy/blob/master/pyproject.toml) file. 

If you don't like to use conda, for vanilla ptypy all you need to do is `pip install .` which will install the core dependencies (numpy, scipy, h5py). 

For fully-fledged ptypy, you can use `pip install .[full]` which will also install some additional dependencies (mpi4py, matplotlib etc.)

The issue mentioned above relating to the version of matplotlib has been fixed already in #555 and will be available with the next release (the fix is already available on the `dev` branch).

The optional cupy dependencies have also been added recently to the https://github.com/ptycho/ptypy/blob/9b73736ae86a5740265230b3cb0324d0dae6ac53/pyproject.toml#L33-L34

so on the `dev` branch, you can also do `pip install .[cupy-cuda12x]` if you would like to install ptypy with cupy support.

We are going to update our documentation soon with more information on how to install ptypy without conda.

PR #584, opened 2024-09-12
Add wavefield preconditioner

jfowkes:
The secret sauce of the LSQ-ML algorithm seems to be what they call average update directions:

image

In reality this is just scaling the ML update directions by 1 / Lipschitz constant for the exit wave error as shown by Andy Maiden et al in "Further improvements to the ptychographical iterative engine". Note that 1/L is known to be the optimal step size for gradient descent on convex functions in the math optimization literature: 
https://math.stackexchange.com/questions/3587312/the-biggest-step-size-with-guaranteed-convergence-for-constant-step-size-gradien

This PR adds an option to ML called "wavefield_precond" that enables this "average update preconditioner". The following results on Moonflower suggests that this is highly effective at accelerating ML as shown in the images below. 

Vanilla ML (all parameters default, 200 iterations):
ML

ML with wavefield_precond=True (all other parameters default, 200 iterations):
ML_precond


ptypy/engines/ML.py
```diff
@@ -290,6 +315,13 @@ def engine_iterate(self, num=1):
             self.pr_grad *= self.scale_p_o
             self.pr_h -= self.pr_grad
 
+            # Lipschitz preconditioner
```
pierrethibault: As just discussed, this application of the preconditioner needs to be moved up before the first application of the smoothing preconditioner (line 257), and needs to be applied once more in a way similar to what is done at lines 308-312.

jfowkes: Unfortunately I can't make it work when the preconditioner is applied twice as suggested (see the commented out code), I get the following:
ml_precond_twice
Are we missing something fundamental here? I think of this more like a damping parameter as in the paper:

> The weightings can be seen as preconditioners of the gradient descent task [19, 21] with small regularization constants δP, δO > 0. These regularized preconditioners result in a behavior similar to the damped least-squares method that penalizes large values in the update, particularly in the weakly illuminated regions.

jfowkes: I tried applying the sqrt of the inverse fluence map twice, but the results are still not as good as before (see the notebook):
ML_wavefield
Something is still not quite right in the code somewhere...

pierrethibault:
You need to take the square root of the sum, not the sum of the square roots.

jfowkes:
> You need to take the square root of the sum, not the sum of the square roots.

How stupid of me, thank you! This now works as expected :)

ptypy/engines/ML.py
jfowkes: Done and working as expected.

jfowkes:
Further testing shows that this "wavefield preconditioner" is also highly effective at accelerating ML on real data, as demonstrated below on the small nanogold dataset from the ptypy tutorials:

Vanilla ML (all parameters default, 200 iterations):
ML

ML with wavefield_precond=True (all other parameters default, 200 iterations):
ML_precond

Really impressive to see this level of improvement on ML without any l2-regulariser or smoothing-preconditioner!

jfowkes:
@daurer to progress this further we'll need a modified `PoUpdateKernel` to compute the fluence maps:
```Python
if self.engine.p.wavefield_precond:
    ob_fln[pod.ob_view] += u.abs2(pod.probe)
    pr_fln[pod.pr_view] += u.abs2(pod.object)
```
This is beyond me I'm afraid... (see line 498 of `ML_serial.py`).

ptypy/engines/ML.py
```diff
@@ -137,6 +153,10 @@ def __init__(self, ptycho_parent, pars=None):
         # Probe gradient
         self.pr_grad_new = None
 
+        # Object and probe fluence maps
+        if self.p.wavefield_precond:
```
daurer: Don't need the if statement here

ptypy/engines/ML.py
```diff
@@ -276,19 +309,30 @@ def engine_iterate(self, num=1):
...
-            # Smoothing preconditioner
-            if self.smooth_gradient:
+            # Smoothing and wavefield preconditioners for the object
+            if self.smooth_gradient and self.p.wavefield_precond:
+                for name, s in self.ob_h.storages.items():
+                    s.data[:] -= self.smooth_gradient(self.ob_grad.storages[name].data / np.sqrt(self.ob_fln.storages[name].data))
+            elif self.p.wavefield_precond:
+                for name, s in self.ob_h.storages.items():
+                    s.data[:] -= self.ob_grad.storages[name].data / np.sqrt(self.ob_fln.storages[name].data)
+            elif self.smooth_gradient:
```
daurer: probably to keep current smooth_gradient switch and just add new wf switch above

ptypy/engines/ML.py
```diff
@@ -377,6 +426,9 @@ def __init__(self, MLengine):
         self.ob = self.engine.ob
         self.ob_grad = self.engine.ob_grad_new
         self.pr_grad = self.engine.pr_grad_new
+        if self.p.wavefield_precond:
```
daurer: no need for the switch here

ptypy/engines/ML.py
```diff
@@ -230,6 +255,14 @@ def engine_iterate(self, num=1):
             else:
                 new_pr_grad.fill(0.)
 
+            # Wavefield preconditioner
+            if self.p.wavefield_precond:
+                self.ob_fln += self.p.wavefield_delta_object
+                self.pr_fln += self.p.wavefield_delta_probe
```
daurer: add this to the divisions below (both cases) - makes it more readible

daurer (changes requested):

jfowkes:
@daurer now we just need the four CUDA kernels `ob_update_ML_wavefield.cu`,  `ob_update2_ML_wavefield.cu`, `pr_update_ML_wavefield.cu`, `pr_update2_ML_wavefield.cu` to compute the object/probe fluences. I would add these based on the non-wavefield ones but I don't understand how the CUDA indexing works in the kernels.

Then I can add some tests for the cupy and pycuda kernels and we should be good to go.

jfowkes:
@daurer the problem with the kernels is the addressing, e.g. for `obj` we do:
```cuda
 const int* oa = addr + 3 + bid * addr_stride;
 obj += oa[0] * H * I + oa[1] * I + oa[2];
 ...
     atomicAdd(&obj[b * I + c], add_val);
```
and we need the equivalent for the fluence map `obj_fln`.

daurer:
@jfowkes yes, came to the same conclusion when looking at the tests. We forgot to increment the fluence arrays in the wavefield kernels. In addition, needed to initialise arrays with zeros() instead of empty().

templates/misc/moonflower_ML_cupy_fft_smoothing.py
```diff
@@ -48,7 +48,7 @@
 # attach a reconstrucion engine
 p.engines = u.Param()
 p.engines.engine00 = u.Param()
-p.engines.engine00.name = 'ML_cupy'
+p.engines.engine00.name = 'ML_serial'
```
jfowkes: @daurer did you mean to modify the cupy template to serial here?

daurer: No, that was just for debugging. Should be reversed.

daurer:
The wavefield preconditioner seems to work across all ML engines (see table below) based on the following parameters

```python
p.engines = u.Param()
p.engines.engine00 = u.Param()
p.engines.engine00.name = 'ML' / 'ML_serial' / 'ML_pycuda' / 'ML_cupy'
p.engines.engine00.numiter = 300
p.engines.engine00.numiter_contiguous = 5
p.engines.engine00.wavefield_precond = True
```

| ML  | ML_serial |
| ------------- | ------------- |
| image | image  |

| ML_cupy  | ML_pycuda |
| ------------- | ------------- |
| image | image  |

daurer (approved):

daurer:
Have also confirmed that it works with both tiled and atomic kernels, i.e.

```python
p.engines.engine00.probe_update_cuda_atomics = False (default) / True
p.engines.engine00.object_update_cuda_atomics = True (default) / False
```

jfowkes:
Many thank @daurer, looking great!

jfowkes:
The failing `ML_serial` engine tests here are not related to this PR, if I raise the `ML_serial` engine test tolerance to `1e-2` which is consistent with the (passing) `ML_cupy` and `ML_pycuda` engine test tolerances then the following tests fail:
```
MLSerialTest::test_ML_serial_all FAILED
MLSerialTest::test_ML_serial_smoothing_regularizer FAILED
MLSerialTest::test_ML_serial_wavefield_preconditioner FAILED
```
This suggests a deeper problem with `ML_serial` but only those parts not used by `ML_cupy` or `ML_pycuda`.

PR #588, opened 2024-11-25
Drop Python <= 3.8 support and add 3.13

ptim0626:
This PR is some housekeeping of Python version support:

- 3.8 reaches EOL so we should drop support
- add 3.13 to the supported Python version
- remove usage of `pkg_resouces` as it is removed from 3.13 (use [`importlib`](https://docs.python.org/3.9/library/importlib.html) and [`packaging`](https://packaging.pypa.io/en/stable/version.html) instead)

I switch over to `conda-forge` and I guess this is probably the better way going forwards given issues with packages in anaconda channel. Let me know if I miss anything.


ptim0626:
Ok Python 3.13 no longer has `pkg_resources`, so it is time to migrate to `importlib`, see https://importlib-resources.readthedocs.io/en/latest/migration.html.

~~WIP~~ Done.

daurer:
Hi Tim, many thanks for this PR. It looks good to me and I have confirmed that accelerated tests are passing ✔️ 
Happy to merge if @bjoernenders is happy as well!?

daurer (approved):

PR #596, opened 2025-01-07
probe from numpy array will not be rescaled to photon count

kahntm:
When a probe was initalized from a numpy array, the _process(ing) applied to that loaded wavefront always scaled it to the number of photons estimated from the loaded diffraction patterns. This happened even when illumiantion.photons was specifically set to None:

`p.scans.scan00.illumination.photons = None`

A similar bug was fixed when loading probes from previous reconstruction (.ptyr) files by specifically setting p.photons to None again. The same "fix" can be used here as well.

The loaded numpy array now remains unchanged. 
**But now it is also impossible to scale a loaded np.array if the users specifically wishes for it.**


kahntm:
Currently the options for illumination.photons are to type **None** (default) or a **number of photons**.  This feature however works only for a probe that is generated from a model and not for probes initialized from a .ptyr file or a numpy array.
image

With those two options (None or a number) it is not possible to clearly extend the scaling functionality to probes initialized from a .ptyr file or a numpy array. I suggest a new accepted string for illumination.photons (something like "maxdiff") to specifically choose to

- None - do not scale the probe, no matter how it is created
- any number - scale the probe to that intensity / number of photons, no matter how it is created
-  "maxdiff" - scale the probe to the intensity of the brightest diffraction pattern, no matter how it is created

image

It is fairly simple to implement and would unify the features functionality across all methods of loading / creating the initial probe.

With "maxdiff" as default any old script that did not specifically set illumination.photons to None to achieve the scaling to the brightest diffraction pattern would still work as before. Only scripts that for some reason did set illumination.photons=None would change their behavior, as in the suggested case, no more scaling would be applied.

Any opinions from the ptypy guardians are welcome.

kahntm:
python 3.8 tests fail when installing dependencies... not in the changed bits of the code

daurer:
> python 3.8 tests fail when installing dependencies... not in the changed bits of the code

Tests should work if you create the PR against dev. Should also male sure your branch is rebased against dev.

daurer:
Make a change to the parameter logic

- Modelled probe: rescale to max diffraction intensity
      - photons = None: rescale to max diff intensity
      - photons - int: rescale to int
      - photons = <0: rescale to max diff intensity
- Provided probe (np array or from .ptyr)
      - photons = None: don't touch my probe
      - photons = int: rescale to int
      - photons = <0: for rescale to max diff intensity

pierrethibault (approved):
Approved. This does not break anything and gives a more natural meaning for `None`.

PR #597, opened 2025-01-24
Renamed engine tests to be picked up by pytest

daurer:


jfowkes:
@daurer looks like you accidentally added the wavefield test in this PR

daurer:
Thanks @jfowkes for the catch. I was indeed just trying to rename the tests so that they are picked up by pytest and github actions. Seems like this is producing some errors, will try to fix those in due course.

daurer:
Fixed issues with smoothing regulariser, base engine tests are now passing locally with scipy 1.14.1 and numpy 2.1.2

Need to investigate why tests are still failing in Github actions.

PR #601, opened 2025-02-13
Control the start of probe update in WASP

### LINKED ISSUE #600: About probe update in WASP algorithm
Dear Contributors,

I found two problems about ptypy WASP algorithm part. 

(1) In the wasp algorithm, there was a parameter called probe_update_start. However, I found that this parameter is not working. even though I set the number to be very high, it still updates the probe at the beginning. Can you help check about this? 
Also, can you check other parameters can work or not, because I also tried the fourier_relax_factor, it seems also make no difference on the results. 

(2) Besides, there seems to be a  little typo on the credits of the WASP. The journal was optics express instead of optica.

Image

ptim0626:
Fix #600 

This PR let users to control the start of probe update by using the parameter `probe_update_start`, as the case for other reconstruction engines. This has not been implemented in WASP previously although it is a valid parameter, and this should be available.

The publication about WASP is also updated (a placeholder was used as the manuscript was still being reviewed when this had been implemented here).

Tests using `probe_update_start` in different WASP engines are added.

### DIFF
--- a/ptypy/custom/WASP.py
+++ b/ptypy/custom/WASP.py
@@ -237,7 +237,8 @@ class WASP(base.PositionCorrectionEngine):
             # update object first, then probe, and accumulate their sum for
             # averaging after going through all the views
             self.object_update(view, ex_old)
-            self.probe_update(view, ex_old, ob_old)
+            if self.p.probe_update_start <= self.curiter:
+                self.probe_update(view, ex_old, ob_old)
 
         # WASP
         self.wasp_averaging()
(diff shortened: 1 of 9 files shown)


ptim0626:
The Python 3.9 failure was because of the change in test introduced in commit 021709f. As noted in the corresponding PR, NumPy > 2 supports float32-type FFT as it produces different result with float64-type FFT.

One solution is to use `scipy.fft` but the result with NumPy FFT does not agree within the normal tolerance. Another solution is to require NumPy > 2 in the dependency files.

daurer (approved):

PR #603, opened 2025-02-28
Fix matplotlib and numpy 2.0 issues

daurer:


daurer:
This has been fixed in #601

PR #604, opened 2025-03-11
Add refractive engines (EPIE, DM, ML)

### LINKED ISSUE #667: Revisit refractive index ML engine
- [x] Do proper comparison of real/imag parts
- [ ] Check code on different examples
- [ ] Consider adding Serial/Cupy engines

jfowkes:
Refractive versions of EPIE, DM and ML (i.e. the object is the refractive index). Just the CPU versions for now (no acceleration).

- Refractive EPIE by @kahntm 
- Refractive DM by @kahntm and @pierrethibault 
- Refractive ML by @jfowkes 

Tested on the following datasets:
- small Siemens stars from @kahntm 
- tungsten blob from @jfowkes 
- nearfield bead from @jcesardasilva

### DIFF
--- a/ptypy/custom/ML_ref.py
+++ b/ptypy/custom/ML_ref.py
@@ -104,6 +104,22 @@ class ML_ref(ML):
     help = How many coefficients to be used in the the linesearch
     doc = choose between the 'quadratic' approximation (default) or 'all'
 
+    [wavefield_precond]
+    default = False
+    type = bool
+    help = Whether to use the wavefield preconditioner
+    doc = This parameter can give faster convergence.
+
+    [wavefield_delta_object]
+    default = 0.1
+    type = float
+    help = Wavefield preconditioner damping constant for the object.
+
+    [wavefield_delta_probe]
+    default = 0.1
+    type = float
+    help = Wavefield preconditioner damping constant for the probe.
+
     """
 
     SUPPORTED_MODELS = [Full, Vanilla, Bragg3dModel, BlockVanilla, BlockFull, GradFull, BlockGradFull]
@@ -166,6 +182,9 @@ class GaussianModel(BaseModel):
         """
         self.ob_grad.fill(0.)
         self.pr_grad.fill(0.)
+        if self.p.wavefield_precond:
+            self.ob_fln.fill(0.)
+            self.pr_fln.fill(0.)
 
         # We need an array for MPI
         LL = np.array([0.])
@@ -207,6 +226,11 @@ class GaussianModel(BaseModel):
                 self.ob_grad[pod.ob_view] += 2. * xi * (1j * np.exp(1j * pod.object) * pod.probe).conj()
                 self.pr_grad[pod.pr_view] += 2. * xi * np.exp(1j * pod.object).conj()
 
+                # Compute fluence maps for object and probe
+                if self.p.wavefield_precond:
+                    self.ob_fln[pod.ob_view] += u.abs2(pod.probe)
+                    self.pr_fln[pod.pr_view] += u.abs2(pod.object)
+
             diff_view.error = LLL
             error_dct[dname] = np.array([0, LLL / np.prod(DI.shape), 0])
             LL += LLL
@@ -214,6 +238,9 @@ class GaussianModel(BaseModel):
         # MPI reduction of gradients
         self.ob_grad.allreduce()
         self.pr_grad.allreduce()
+        if self.p.wavefield_precond:
+            self.ob_fln.allreduce()
+            self.pr_fln.allreduce()
         parallel.allreduce(LL)
 
         # Object regularizer
--- a/ptypy/custom/ML_ref.py
+++ b/ptypy/custom/ML_ref.py
@@ -1,0 +1,383 @@
+# -*- coding: utf-8 -*-
+"""
+Maximum Likelihood reconstruction engine (refractive index).
+
+TODO.
+
+  * Implement other regularizers
+
+This file is part of the PTYPY package.
+
+    :copyright: Copyright 2014 by the PTYPY team, see AUTHORS.
+    :license: see LICENSE for details.
+"""
+import numpy as np
+
+from ptypy import utils as u
+from ptypy.utils.verbose import logger
+from ptypy.utils import parallel
+from ptypy.engines import register
+from ptypy.engines.ML import ML, BaseModel
+from ptypy.core.manager import Full, Vanilla, Bragg3dModel, BlockVanilla, BlockFull, GradFull, BlockGradFull
+
+
+__all__ = ['ML_ref']
+
+
+@register()
+class ML_ref(ML):
+    """
+    Maximum likelihood reconstruction engine (refractive index).
+
+
+    Defaults:
+
+    [name]
+    default = ML_ref
+    type = str
+    help =
+    doc =
+
+    [ML_type]
+    default = 'gaussian'
+    type = str
+    help = Likelihood model
+    choices = ['gaussian','poisson','euclid']
+    doc = One of ‘gaussian’, poisson’ or ‘euclid’. Only 'gaussian' is implemented.
+
+    [floating_intensities]
+    default = False
+    type = bool
+    help = Adaptive diffraction pattern rescaling
+    doc = If True, allow for adaptative rescaling of the diffraction pattern intensities (to correct for incident beam intensity fluctuations).
+
+    [intensity_renormalization]
+    default = 1.
+    type = float
+    lowlim = 0.0
+    help = Rescales the intensities so they can be interpreted as Poisson counts.
+
+    [reg_del2]
+    default = False
+    type = bool
+    help = Whether to use a Gaussian prior (smoothing) regularizer
+
+    [reg_del2_amplitude]
+    default = .01
+    type = float
+    lowlim = 0.0
+    help = Amplitude of the Gaussian prior if used
+
+    [smooth_gradient]
+    default = 0.0
+    type = float
+    help = Smoothing preconditioner
+    doc = Sigma for gaussian filter (turned off if 0.)
+
+    [smooth_gradient_decay]
+    default = 0.
+    type = float
+    help = Decay rate for smoothing preconditioner
+    doc = Sigma for gaussian filter will reduce exponentially at this rate
+
+    [scale_precond]
+    default = False
+    type = bool
+    help = Whether to use the object/probe scaling preconditioner
+    doc = This parameter can give faster convergence for weakly scattering samples.
+
+    [scale_probe_object]
+    default = 1.
+    type = float
+    lowlim = 0.0
+    help = Relative scale of probe to object
+
+    [probe_update_start]
+    default = 2
+    type = int
+    lowlim = 0
+    help = Number of iterations before probe update starts
+
+    [poly_line_coeffs]
+    default = quadratic
+    type = str
+    help = How many coefficients to be used in the the linesearch
+    doc = choose between the 'quadratic' approximation (default) or 'all'
+
+    """
+
+    SUPPORTED_MODELS = [Full, Vanilla, Bragg3dModel, BlockVanilla, BlockFull, GradFull, BlockGradFull]
+
+    def __init__(self, ptycho_parent, pars=None):
+        """
+        Maximum likelihood reconstruction engine (refractive index).
+        """
+        super().__init__(ptycho_parent, pars)
+
+    def _initialize_model(self):
+
+        # Create noise model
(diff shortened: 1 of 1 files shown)


kahntm:
Hi. Thanks for implementing this. I will have a look at it soon.
As for "good" data to test it with: Any dataset should do. But some where the phase wraps at least once or twice would be the best.

jfowkes:
I have now added the wavefield preconditioner to refractive index ML (option `wavefield_precond = True`).

jfowkes:
Here is a comparison of ML with the smoothing preconditioner:
ML
and refractive-ML with the smoothing preconditioner:
ML_ref
Seems to work okay on this easy dataset.

jcesardasilva:
Thank you for sharing this with us. It seems to work, but I am a bit concerned about the gray values of the amplitude. I assume that both calculations were done on the same dataset, am I right? However, in the results obtained with "ML with the smoothing preconditioner," the range of gray values for the amplitude goes from 0.5e-10 up to 0.79e-10 (a delta of 2.9e-11). In contrast, the results from "refractive-ML with the smoothing preconditioner" show a range from 0 to 2.8 (a delta of 2.8). Additionally, the contrast seems inverted for the amplitude images: what is darker in one appears lighter in the other, and vice versa. 

Could a minus sign be floating somewhere where it's not supposed to be? 

Additionally, the probes have completely different power values, with one P = 8.5e8 and the other having P =4.8e8, which is almost a factor of 2 different. This obviously also affects the range of the color scale. This may indicate that something not so correct occurred, as even if one is now retrieving the refractive index for the object rather than the exponential of the phase, the probe should have remained very similar. 

Could one force the amplitude to be between 0 and 1 (or even 1.2 to allow some flexibility) to check if there are improvements?

jcesardasilva:
Is this also implemented for the near-field ptycho? If so, is it possible to try it in the dataset I send you @jfowkes?  The probe retrieval may complicate things a bit, though.

jfowkes:
@jcesardasilva yeah sorry I’m being lazy with the plotting (as ptypy isn’t setup for plotting refractive indexes) so for the refractive ML it’s actually plotting the amplitude and phase of the refractive index rather than the real and imaginary part, I’ll see if I can hack the plot script to fix this. It should work with nearfield data out of the box so I will try with your dataset today.

daurer:
As discussed at the developer meeting, we should do a proper comparison of the results (not using ptypy.utils) so that we can discuss whether the refractive ML code makes sense. @jfowkes do you still have the scripts and could do this?

jfowkes:
Yes I still have the runscripts, happy to do the comparison properly. I should add that it does not work for the second example as the gradients explode.

jfowkes:
As requested, here is a ML vs refractive ML comparison on Maik's small Siemen's stars dataset.

**ML**
Here we plot the amplitude and phase of the object
ML

**Refractive ML**
Here we plot the real and imaginary parts of the refractive index (real=phase, imag=amplitude)
ML_ref

@kahntm @jcesardasilva let us know what you think. Can also supply the ptyrs for detailed analysis.

ltang320:
Quick question: the refractive ML result looks as though it’s been blurred by convolution with a kernel. Was that intentional?

jcesardasilva:
Hi @jfowkes 

The results look very good. 

I noticed the **Refractive ML** results look smoother, as if they were strongly regularized (TV reg., maybe?). Is that the result of regularization? A good test would be to calculate the Fourier Ring Correlation between the **normal ML** and **Refractive ML** results to see which frequencies they correlate up to. Or, if you have two half-datasets, you could calculate the resolution from the **ML** results and compare it with the resolution from **Refractive ML**. I also noticed that the `P`-value above the probe gives different values, which does not seem to be a big problem, but do you know why does this happen? 

About the test data, I do have one, but it comes from Near-field ptychography and goes over $\pi$ in the phase gray level. So it's a tricky one, particularly because the low frequencies converge slowly, but **DM** handles it quite well, and it is a good opportunity to check how well **Refractive ML** performs on wrapped phases, which is the whole goal of that approach. Here is a preview of what to expect as results (before unwrapping):
image

Sorry, it is upside down because I displayed it with matplotlib using `origin='lower'`. 

If you want to tackle that data, it is at the cloud link: https://hub.neel.cnrs.fr/index.php/s/En8zXFFicbzfe38
The data is in the file `bead_NFP_070nm_0000.h5`.
In the same folder, I included the Ptypy script to run it with the proper parameters. 
Do you want to try it?

jfowkes:
> Quick question: the refractive ML result looks as though it’s been blurred by convolution with a kernel. Was that intentional?

@ltang320 not intentional, I'm running with the gradient smoother for both ML and ML_ref so it may be due to that:
```Python
p.engines.engine00.smooth_gradient = 20.
p.engines.engine00.smooth_gradient_decay = 1/50.
```
I can test without this option for `ML_ref` and see if it's less blurry. Can't think what else it could be.

jfowkes:
Thanks @jcesardasilva, I'm currently trying with the nearfield dataset `v97_h_nfptomo_15nm_subtomo001_0000.h5` that you gave me last time. Many thanks for the new dataset, looks great and as you say tests how well refractive ML performs on phase wrapped datasets, so will definitely give that one a go as well!

For the refractive ML reconstruction on the Siemens stars dataset, I think the blurring might be due to the gradient smoother but I will investigate this. Good idea on the FRC, should be able to use the code that you and Pierre supplied for PtyPy (can't remember if it was merged or if it's still in a PR here).

Regarding the different Power values on the probe, I think that's because this is a degree of freedom in the reconstruction and we're not constraining it in any way. So the probe power can increase a bit and the object power decrease a bit but when they're multiplied together the exit wave stays the same as they cancel out.

ltang320:
Hi, @jfowkes . That makes sense to me. Thanks for looking into it.

jfowkes:
**Refractive ML (no gradient smoothing)**
Here we plot the real and imaginary parts of the refractive index (real=phase, imag=amplitude)
ML_ref_no_gradient_smoothing

Here is the reconstruction using refractive ML without the gradient smoother, seems to get rid of the blurring.

jcesardasilva:
It is very good. It is great to see the improvement it makes to the data, particularly in the amplitude-contrast image. I would love to see the phase in another colormap, because that "cyan" color does not seem to enhance the contrast between the support and the sample. But this is a tiny thing. The important thing is to see that the information is there.

jfowkes:
@jcesardasilva agreed the cyan is not nice, I'm using ptypy's default imaginary part colormap. What colormap would you like? This should be very easy to change in the plotting code.

jcesardasilva:
I don't have a favorite, but I want one that shows the contrast better. Maybe the same as the amplitude image, or something very pale, like "bone". Or even just adding an offset to move away from the very light colors.

jfowkes:
**Refractive ML (no gradient smoothing, bone colormap)**
ML_ref_no_gradient_smoothing_bone

@jcesardasilva how's that? Using the bone colormap here.

jcesardasilva:
Great. Thank you very much. 

It seems the phase shift values are lower for the object in the new reconstruction than in the first trials. For example, with standard ML, we had the pink values (above zero, so positive) in the sample support and close to -2 (yellow) in the object. A range of approximately 2 rad for the object. 

In the latest reconstruction, it appears we have values close to 0.55 in the substrate and -0.6 for the object, which is a bit more than a 1 rad shift.

I don't have information on the sample's expected phase shift, but since the amplitude has improved, I am wondering what quantitative values are expected.

daurer:
Need to be careful here when comparing the images. With refractive ML the real part (which is plotted on the left) should be compared with the phase (which is plotted on the right).

Should probably switch the images in the new real/imag plotting and display imaginay part on left and real part on the right.

jfowkes:
Yeah good call, I will switch the image positions and maybe add better labels.

jfowkes:
@jcesardasilva I've switched the plots for refractive ML so you should now be able to do a direct comparison

**Normal ML**
ML

**Refractive ML**
ML_ref

jcesardasilva:
Great! It seems more in line with, and more coherent with, the values obtained with standard ML. 

As @daurer already pointed out, we no longer display the **amplitude** and **phase** contrast images, but the real ($1-\delta$) and imaginary part ($\beta$) of the refractive index. If I am not misunderstanding the results, I noticed you plotted the $1- \delta$ part from $-\pi$ to $\pi$, but this is not phase anymore. 

Maybe I am being too meticulous, without wanting to be pedantic, but $\delta$ values range from $10^{-7}$ to $10^{-5}$ in X-rays. I don't know whether you are showing $1-\delta$ or simply $\delta$, but the scale looks more like phase shifts, from $-\pi$ to $\pi$. Since we are measuring refractive index, shouldn't we expect values within the typical range of refractive-index decrement values? 

Concerning the colormap choice for display, please choose the one you feel suits it best. I had suggested `bone`, but I know some people prefer `jet` or `viridis` for the **refractive index decrement**. I don't have a strong preference for any one.

jfowkes:
Thanks @jcesardasilva, my understanding is that we're plotting what in the following is called  $\mathcal{I}(\tilde{O}_r)$ and $\mathcal{R}(\tilde{O}_r)$. Hopefully the math below makes it clearer to understand?

Screenshot 2026-09-16 at 10 02 45

jcesardasilva:
Ok, thank you for this. 
So, if I understand correctly, for an energy of, let's say, 8 keV, and supposing a sample with $\delta=10^{-6}$ and a thickness of $10\mu m$, we have: 
```
In [1]: import numpy as np

In [2]: wavelen = 12.4*1e-10/8

In [3]: (2*np.pi/wavelen)*(1e-6*10e-6)
Out[3]: 0.4053667940115862
```
Thus, the $\delta$-part is indeed close to unity, instead of having 6 figures after the comma. It makes sense. Thanks.

jfowkes:
Here are the results for @kahntm's tungsten blob dataset, I'm running 500 iterations of EPIE then 500 of ML. The refractive engines seem to do a good job of getting rid of the phase vortices in the blob.

**Normal EPIE + Normal ML**
EPIE_ML

**Refractive EPIE + Refractive ML**
EPIE_ref_ML_ref

@jcesardasilva I will try out your nearfield bead dataset next.

jcesardasilva:
This is amazing. 
It seems to me that there is still one pi-wrapping around the particle borders? Is it only one impression? The colors go from mild gray outside the particle to almost white at the border, then to a more grayish color in the middle of the particle.

I know my reasoning appears weird, since you are now taking the real part of the function to extract the image. So, it is confusing for me to see value very close to $\pi$ at the particle border.

One way to check is to extract a 1D gray-level profile across the particles to verify this. If the transition from air to sample does not show any sharp jumps, then it was just my impression.

jfowkes:
Yeah just discussed this with @daurer and it does appear there's still some phase wrapping around the particle border, as you say a small strip close to $-\pi$. Ptypy's default plotting for the right image hardcodes the range from $-\pi$ to $\pi$ which probably isn't what we want. Also we should probably be plotting the exponential of the left image, since the maths says the imaginary part corresponds to $-ln(A_r)$ and really we want to see the absorption $A_r$.

jfowkes:
@jcesardasilva finished running DM+ML on your bead dataset with and without refractive. Here are the results.

**Normal DM + Normal ML**
DM_ML

**Refractive DM + Refractive ML**
DM_ref_ML_ref

jcesardasilva:
Thank you very much. 
First, we see that **refractive ML** works, so far so good! 

Nevertheless, the challenge here is the sample part and the near-field regime, which is somewhat disconnected from refractive ML. 

Just for completeness, let me explain the case.  As you can see, the algorithm did not converge with "only" 2000 iterations, which also happens with normal DM+ML. A large halo artifact appears around the sample, which is not physical but typical in the near-field regime. The point is that phase wraps will only start appearing once the halo is attenuated (and then the refractive ML can be fully tested), but to do that, we need many more DM iterations. To illustrate the problem, the sample is about 70 microns thick, the energy was 17.05 keV, which gives:

```
In [1]: E = 17.05

In [2]: wavelen = 12.4e-10/E

In [3]: import numpy as np

In [4]: (2*np.pi/wavelen)*(1e-6*70e-6)
Out[4]: 6.047565858160351
```
We would then expect the gray level to change by at least a value 6 in the sample, relative to the air region, but you can see the range of change in gray level is less than 1. 
As you can also see in the probe, we barely have any low-frequency information, only high-frequency. While the probe is not fully developed, the halo won't disappear, and the low-frequency in the sample won't be right. 

So, ok, **refractive ML** works, but we are still far from convergence for that specific case, which is largely independent of **refractive ML** itself. That is exactly where I hope **L-BFGS** could help, since it's disconnected from this issue.

PR #620, opened 2025-07-02
HDF5Loader: Add option to provide defocus for nearfield data

daurer:
For nearfield data, it can be convenient to provide a defocus value to the data loader and calculate the effective detector distance and pixelsize as part of the preparations in the Hdf5Loader class.

### DIFF
--- a/ptypy/experiment/hdf5_loader.py
+++ b/ptypy/experiment/hdf5_loader.py
@@ -320,6 +320,12 @@ class Hdf5Loader(PtyScan):
     type = list, ndarray
     help = This is the array or list with the re-ordered indices.
 
+    [nearfield_defocus]
+    default = None
+    type = float
+    help = Distance from sample to focus (for nearfield only)
+    doc = If set, magnification will be calculated automatically and applied to detector distance and pixelsize 
+
     """
 
     def __init__(self, pars=None, swmr=False, **kwargs):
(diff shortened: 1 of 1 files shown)


jcesardasilva:
Thanks. Indeed, it can be convenient to do that. However, you then need additional parameters, such as z1 (dist focus-to-sample) and z2 (distance sample-to-detector), with which one can calculate z12 and the z_eff (scaled by the Fresnel scaling theorem). And if the pixel size given is the physical pixel of the detector, one needs an additional routine to calculate the demagnification and, finally, the final pixel size.

daurer:
Thanks for the comment @jcesardasilva. The focus-to-sample distance (z1) is this new parameter that I named `nearfield_defocus` - maybe not the best naming I guess. Should we call it `nearfield_focus_to_sample` instead?
The sample-to-detector distance (z1) is implicitely already given by the regular distance parameter from PtyScan, here interpreted as the focos-to-detector distance (z1+z2). With that information, the magnification is calculated as well as the effective sample-to-detector distance and the effective pixel size (see lines 608-610).

I have tested this on real nearfield data from I14 and it seems to work. But if you could double check my implementation of the rescaling based on Fresenel theorem and confirm its correctness that would be great!

bjoernenders (approved):

PR #633, opened 2025-09-15
Final report implementation

### LINKED ISSUE #585: final_report
During the ptypy workshop we came up with the idea that ptypy should give a final report at the end of a reconstruction. Some metrics that are worth knowing about the reconstruction:

- what was the probe size (probes are messy, so a metric or various metrics would need to be found and reported)
- what was the average step size
- using those numbers, what was the (linear / area) overlap in the sample plane
- using those numbers, what was the oversampling in the detector plane
- what do these photon error metrics mean
- maps of which pixels are covered by how many views (how often do they get updated)
- fluence map - how much of the total intensity ended up in each pixel of the object
- just again saying there were X diffraction patterns of size N * M

Some of these things could be printed on the terminal at the end of the reconstruction.
All of these things could be added within the .ptyr file.

There was also a wish to have an FRC function implemented in the utils

pierrethibault:
Implemented basic reporting structure (report will appear in ptyr file).

Next steps: 

- [x] implement function that evaluates probe area and linear size
- [x] function that calculate average step size
- [x] linear overlap
- [x] area overlap
- [x] oversampling
- [x] fluence map
- [x] coverage map
- [x] transmission map (multiplied with the object transmission)
- [x] summary of other existing metrics (field of view, number of scan points, ...)

All this information is stored in a dictionary called Ptycho.report, and saved in the ptyd file under content/report. We also need a utility that formats the information for easy consultation

- [ ] markdown template
- [ ] utility to generate the report and figures

This closes #585


.idea/codeStyleSettings.xml
daurer, line 1: this should be remove

PR #606, opened 2025-03-27
Add a routine for Fourier Ring correlation to the new `metrics.py` in `utils`

jcesardasilva:
This PR is a draft to foster the discussion about the inclusion of the `FRC` in `metrics.py`

The list of things to discuss and improve is in the issue #605

Issue #605, opened 2025-03-20
Add a routine for Fourier Ring correlation to the new `metrics.py` in `utils`


jcesardasilva:
Create the `metrics.py`in `utils` and add a routine for Fourier Ring Correlation (FRC).

jcesardasilva:
The first draft of the FRC routine has been included in `metrics.py` in `utils.` The commit is in the `cnrs` branch: 0e88137ab2d815cfd58c021914d95c278ba87dfe
The implementation is based on the following publication: [van Heel & Schatz 2005](https://doi.org/10.1016/j.jsb.2005.05.009)

Since this is only for comparing 2D images, I have only implemented the 1-bit threshold. The half-bit threshold must only be used for tomography, where we divide the same dataset in two. 

Keep in mind that although it seems to be working, this is a work in progress. I put it here so you can contribute to improving it if you want. The implementation is deeply inspired by the FSC implementation in `Toupy`, a Python package for tomography for which I am responsible for the development. 
The documentation for `Toupy` is here: [FSC in Toupy](https://toupy.readthedocs.io/en/latest/rst/toupy.resolution.html)
The source code for `Toupy` is here: [FSC in Toupy Github](https://github.com/jcesardasilva/toupy/tree/master/toupy/resolution)

I think it would be nice to try to implement it with more differences relative to the `Toupy` implementation (where the FSC is a python `Class`) and closer to `ptypy`. Please let me know what you think about it. 

Here is the TODO list of possible improvements:


- [ ] - the names of the functions and variables to be in full harmony with the `ptypy` style. 

- [ ] - Image registration: The two images must be aligned before the FRC is computed. Otherwise, misaligned images can distort the results. In the current implementation, I am using the sub-pixel registration routines from the package `Scikit-image`: [phase_cross_correlation](https://scikit-image.org/docs/0.23.x/auto_examples/registration/plot_register_translation.html#sphx-glr-auto-examples-registration-plot-register-translation-py). This would add an extra dependency to `ptypy`. Therefore, we could consider implementing a routine for image registration and avoid extra dependency (or not!). What do you think? 

- [ ] - For the FRC calculation, I need extra array manipulation. For this, I created the additional functions: `nyquist`, `ringthickness`, `apodization`, and `imgregistration`. I personally think they would fit better in the `array_utils.py`. Please, let me know your view on this point. 

- [ ] - For the FRC, we need to plot some things. Should the plotting parts be handled py the `plot_utils.py`

[...]

PR #634, opened 2025-09-16
Fourier Ring/Shell Correlation module 

daurer:
This is a clean starting point based on the initial draft #606 for FRC calculation from @jcesardasilva  and additional code added for FSC calculation by @Letizia97 (see also #605)

TODO

- [x] Homogenize `metrics.py` with two main FRC/FSC functions at the top and all utiliy functions below (@jcesardasilva )
- [x] Make sure it works for both 2d and 3d case (@jcesardasilva )
- [x] Try to merge utilitiy functions that are used for both 2D and 3D (@jcesardasilva )
- [ ] Upload custom 2d alignment code to replace scikit-image (@gfev )
- [ ] Automated intersection determination based on 1-bit and 0.5 bit thresholds (@gdev) 
- [ ] Create a tutorial notebook that uses the new module and generates nice plots (@Letizia97 @jcesardasilva )
- [ ] Looks into performance optimsation (@pierrethibault )


jcesardasilva:
I noticed now that there are two types of apodization in 3D:
1) A square-like Hanning window for the transverse cross-section of the tomogram and one circle-like for the axial cross-sections, given that the information outside that circle in tomography is basically noise or artifacts. 
2) A square-like Hanning window in all three dimensions (a cubic-like apodization window function if you prefer). So, this ignores that, in the tomography, the information that matters in each axial cross-section is only inside a circle. 

Do we need to keep both? If so, we need to implement three different kinds of apodization, the two I mentioned above and the one for the 2D images only.

ptypy/utils/metrics.py
```diff
@@ -0,0 +1,544 @@
...
+        result = np.outer(window1D1, window1D2)
+    
+    elif len(inputarray.shape)==3:
+        ns, nr, nc = inputarray.shape         
+        
+        window1D1 = hanning_like_window(ns, apod_width)
+        window1D2 = hanning_like_window(nr, apod_width)
+        window1D3 = hanning_like_window(nc, apod_width)
+
+        windowaxial = np.outer(window1D2, window1D3)
```
pierrethibault: Why not use two nested outer products here?

ptypy/utils/metrics.py
```diff
@@ -0,0 +1,544 @@
...
+    input2 = imgregistration(input1,input2,upsamp=100)
+    
+    # Forcing to using 1 bit threshold because it is ring correlation
+    # 1/2 bit threshold must only be used for tomography
+    print('Computing FRC using 1 bit threshold')
+    snrt = 0.5
+    
+    # Apodization of the borders
+    window = apodization(input1, apod_width)
+    img1_apod = input1 * window
```
pierrethibault: The input images probably aren't "whitened". Here, multiplying with a window sends the edge values to 0. I think that artifacts would be reduced if we use instead `img1_apod = window * (input1 - input1.mean()) + input1.mean()`.

ptypy/utils/metrics.py
```diff
@@ -0,0 +1,544 @@
...
+    
+    # initializing variables
+    C = np.empty_like(f)
+    C1 = np.empty_like(f)
+    C2 = np.empty_like(f)
+    npts = np.zeros_like(f)
+    
+    print("Calculating the correlation...")
+    index = ringthickness(F1) # indexes for the ring thickness
+    for ii in range(len(f)):
```
pierrethibault: There is a way to make this faster using `numpy`'s `digitize` and `bincount`. I have an example code I can show you.

ptypy/utils/metrics.py
jcesardasilva: Maybe I did it wrong, but in my case, it did not work. Could you make a quick draft of what you meant, please?

ptypy/utils/metrics.py
jcesardasilva: Supposing I understood your remark well, the idea of the implemented apodization here was to transition slowly to 0 at the edges, thereby avoiding a sharp transition. If the edge values are not truly zero, given that the discrete FFTs produce numerous replicas around, you may end up with the classical `(input) x (rect function)`, leading to artifacts when implementing the discrete FFTs.

ptypy/utils/metrics.py
jcesardasilva: I'd be happy to review your example code, which could save me time if I were to do it myself. Does the bincount work in both 2D and 3D arrays?

jcesardasilva:
I made a couple of changes in the apodization functions, given their importance. You can see the commit 22e8129ff03d326d94d4cc676c6617199be52258
Tomorrow, I will keep working on the functions. 

I think we could merge `fourieringcorrelation` and `fouriershellcorrelation`, giving the capabilities to the function to identify if the images are 2D or 3D. Except if you want to keep them separated. Please let me know.

jcesardasilva:
Just in case this can be of interest, I've just found out that `scikit-image` has a very nice and efficient window function that works in both 2D and 3D without problem: https://github.com/scikit-image/scikit-image/blob/v0.25.2/skimage/filters/_window.py#L10-L131
But of course, this would add an extra dependency.

jcesardasilva:
There are now two functions, `frc` for the 2D FRC, and `frc` for the 3D FSC. They are shorter and a bit less repetitive thanks to auxiliary functions. 
I have not removed the alignment routine yet, leaving the task for someone else who has access to @pierrethibault 's custom code.  Commit dfcd35d679725fde26b08caa65d7c4b7c628830a
I also included changes in the Jupyter notebook, showing that both `frc` and `fsc` seem to be working locally. I still need a tomographic dataset to test the `frc`, but I need to wait for the unarchiving of my data. If you have some available, please go ahead with the test on your data, and I can help if needed.

ptypy/utils/metrics.py
pierrethibault: My point is that sending the signal to 0 actually can introduce artifacts if the average signal is everywhere far from 0, especially if the apodization width is small. I tried to make a drawing to explain that...

image

jcesardasilva: Great, thanks. Now I see what you meant. Well, the apodization steps are now coded in independent auxiliary functions. Therefore, we can easily change them to play with different approaches without having to rewrite the main code. One could try to do what you said.

jcesardasilva: Just one question: you mentioned earlier the idea of doing `img1_apod = window * (input1 - input1.mean()) + input1.mean()`. But, I noticed you still use the function `window`, which will bring the values to zero at the borders. If I understand correctly, we are supposed to remove the `window` from the multiplication, and do only `img1_apod = (input1 - input1.mean()) + input1.mean()`, right? Or would you happen to have another window function in mind?  Or, by window, you mean the natural windowing (selecting the region within the Nyquist frequency range) we do when performing the discrete FFTs ?

daurer:
Just a few comments for cleaning up at the end. We should move any example code/notebooks into the ` templates/notebooks` folder and remove the extra stuff that has been added to .gitignore

ptypy/utils/metrics.py
```diff
@@ -0,0 +1,595 @@
...
+    plt.plot(fn, FRC.real, "-b", label="FRC")
+    plt.plot(fn, T, "--r", label="1 bit threshold")
+    plt.legend()
+    plt.xlim(0, 1)
+    plt.ylim(0, 1.1)
+    plt.xlabel("Spatial frequency/Nyquist [normalized units]")
+    plt.ylabel("Magnitude [normalized units]")
+    plt.show()
+    plt.savefig(plot_name)
+    return None
```
daurer: I think we should remove this from the module, can instead show how to plot in the example notebooks. Should also remove the matplotlib import consequently

jcesardasilva: I've just removed the plotting part and the matplotlib import.

jcesardasilva:
I removed the dependence on matplotlib and the plotting section. I also adjusted the docstring style in `compute_intersection` to match the other docstrings. 

One crucial point is that the user needs the pixel value so that dividing it by the intersection value allows for an estimate of the spatial resolution. At any point in the FSC/FRC calculations, we are inputting the pixel size. Will `PtyPy` provide the pixel size for the FSC/FRC calculation at any stage, or will it be up to the user to determine it? In any case, it is worth showing in the tutorial that one needs the pixel size to estimate the resolution based on the intersection value.

daurer:
This might be ready for merge, but testing notebooks should be moved to templates folder removed

daurer:
Minimal version of FRC/FSC is ready to be merged. Need to clean up example notebooks and move them to right place. Consider open new issues for outstanding work (alignment, apodization).

PR #638, opened 2025-09-21
Clipping magnitude and phase 

yurirtonin:
The clipping input is now assumed to be either a 2-tuple or 4-tuple. 

In case it is as 2-tuple, it is assumed to be (mininum_magnitude, maximum_magnitude)
In case it is as 4-tuple, it is assumed to be (mininum_magnitude, maximum_magnitude, minimum_phase,maximum_phase)

For ignoring a specific clipping, use None. For instance: (minimum_magnitude, None, None, maximum_phase)

I kept the approach that was already implemented for the magnitude clipping, so it is done right after object update and before probe update.

Done:
- [x] Change clipping functions to clip not only magnitude but also phase (CPU and cupy)
- [x] Renamed .cu file and functions that referred to "clip magnitude" to "clip object" 
- [x] Add clipping functionality to stochastic engine (CPU and cupy)
- [x] Add clipping functionality to ML engine (CPU and cupy)
- [x] Edit WASP engines to also use the new clipping strategy

Still possible ToDos:
- [ ]  add clipping strategy to pycuda engines
- [ ] Improve clipping strategy (see comment below)

One open point of discussion: what is the "correct" way of doing such a clipping? Currently, we clip the magnitudes (np.abs) followed by clipping of the phase (np.angle), which would result in a behavior in the complex plane as the following:

Image

### DIFF
--- a/ptypy/custom/WASP_serial.py
+++ b/ptypy/custom/WASP_serial.py
@@ -328,15 +328,7 @@ class WASP_serial(WASP):
                 self.benchmark.wasp_averaging += time.time() - t1
                 self.benchmark.calls_wasp_averaging += 1
 
-                # Clip object (This call takes like one ms. Not time critical)
-                if self.p.clip_object is not None:
-                    clip_min, clip_max = self.p.clip_object
-                    ampl_obj = np.abs(ob)
-                    phase_obj = np.exp(1j * np.angle(ob))
-                    too_high = (ampl_obj > clip_max)
-                    too_low = (ampl_obj < clip_min)
-                    ob[too_high] = clip_max * phase_obj[too_high]
-                    ob[too_low] = clip_min * phase_obj[too_low]
+                self.clip_object(ob)
 
             # Re-center the probe
             self.center_probe()
--- a/ptypy/custom/WASP_serial.py
+++ b/ptypy/custom/WASP_serial.py
@@ -328,7 +328,7 @@ class WASP_serial(WASP):
                 self.benchmark.wasp_averaging += time.time() - t1
                 self.benchmark.calls_wasp_averaging += 1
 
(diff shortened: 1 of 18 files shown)


yurirtonin:
Fixed a couple of refactoring I had missed. Nonetheless, the latest test error on WASP_serial regarding the complex type of the object I am not sure how to solve yet. Running the WASP and WASP_cupu engines directly do not present this error

ptypy/custom/WASP_serial.py
```diff
@@ -328,15 +328,7 @@ def engine_iterate(self, num=1):
...
-                # Clip object (This call takes like one ms. Not time critical)
-                if self.p.clip_object is not None:
-                    clip_min, clip_max = self.p.clip_object
-                    ampl_obj = np.abs(ob)
-                    phase_obj = np.exp(1j * np.angle(ob))
-                    too_high = (ampl_obj > clip_max)
-                    too_low = (ampl_obj < clip_min)
-                    ob[too_high] = clip_max * phase_obj[too_high]
-                    ob[too_low] = clip_min * phase_obj[too_low]
+                self.clip_object(ob)
```
daurer: @yurirtonin could be wrong but it looks like you are passing a numpy array where as the function expects a Container?

Maybe this explains the funny error in the WASP tests?

yurirtonin: Hi Benedikt, indeed. Problem solved! Thanks once again

daurer:
Merge, but keep an issue for the implementtation of different strategies for amplitude + phase clipping (including the graph)

daurer:
Add warning for not-availablilty in pycuda engines

daurer (approved):

PR #639, opened 2025-09-22
Quality-guided phase unwrapping

pierrethibault:
As discussed at the end of the Hackathon 2025, I am adding here my implementation of the quality-quided phase unwrapping method.

This is manifestly not ready to pull, because I still need to figure out how this will be integrated in the installation. It looks like there are two main options:
1. Have a `build_ext` that compiles the file, and use a wrapper with `ctypes`. For now the result looks messy, because compiled libraries have different names on different platforms and some code must be added to search for the shared library.
2. Wrap the c file into a real python extension, which seems to be better handled at installation, but requires a bit more of work.

Currently I am leaning towards option 2, which will give a cleaner final result, but I know that Benedikt @daurer suggested using ctypes. Waiting for input!

Todo:

- [x] Wrap the function
- [x] Setup install script and metadata
- [x] Add some tests
- [ ] Possible improvements: allow for more than one seed for the region-growth process (this requires passing a list of coordinates, which would be easier to handle if the code is converted to a real python Extension (option 2 above).

### DIFF
--- a/ptypy/utils/unwrap/_qmunwrap.c
+++ b/ptypy/utils/unwrap/_qmunwrap.c
@@ -1,0 +1,283 @@
+/*
+Quantised-level quality-map phase unwrapping.
+
+This implementation is meant to accelerate the quality map approach
+through a binning process of the quality map. This eliminates the need for sorting
+at the price of a (low-risk) non-sequential unwrapping.
+
+Author: Pierre Thibault
+Date: First version sometimes around 2010.
+*/
+
+#include "math.h"
+#include "stdlib.h"
+#include "stdio.h"
+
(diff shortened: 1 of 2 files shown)


daurer:
@pierrethibault I agree that wrapping this into a real Python extension is the better option!

pierrethibault:
The current suggested implementation has reinstated a `setup.py` file. There is a way to avoid this, but then `c-compiler` must be added as dependency @bjoernenders what do you think?

bjoernenders:
c-compiler is thumbs up. Please go ahead

daurer (approved):

PR #644, opened 2026-01-07
Improve ML cupy memory estimation

daurer:
In some cases, we have noticed that when the last block of data is allocated in the ML cupy engine, there is not enough memory available despite upfront calculations on how much device memory is free. 

The origin of this issue could be related to discrepancies between memory allocation of cupy and via the lower level CUDA runtime API - this needs further investigation.

As an intermediate solution, I am adding the capabilities to leave more extra device memory (now defined as a fraction of total memory) and allow users to change this "safety margin" by adding an environment variable PTYPY_DEVICE_MEM_SAFETY. 

The new default for this safety margin is 2.5% of the total device memory or at least 400 MB (previously it was hardcoded to 200 MB) to make sure we can always operate on low-memory devices.


daurer:
Keep fractional idea but never go below an absolute value (increase that from 200 MB to 400 MB)

